PR #26, opened 2015-06-13
Features/gradlfication

crankycoder:
This converts the build to use gradle instead of Maven to do the build.

I've added a dependency to com.jakewharton.sdkmanager so that you no longer have to run shell scripts to setup your enviroment.  This should mean that as long as you've got a working Java enviroment, all the Android SDK and support library dependencies should get installed automatically for you in ~/.android-sdk

All test cases have been dropped because of robolectric 2.4 compatibility problems.  Resuscitating them from source control ought to be fairly straight forward, but they'll all have to be updated to robolectric 3.0 anyway.


ecgreb:
Thanks @crankycoder !

I'm going to make an attempt today to upgrade Robolectric and get the unit tests up and running again before we merge this in. Stay tuned.

lost-sample/build.gradle
```diff
@@ -0,0 +1,49 @@
+buildscript {
+    repositories {
+        jcenter()
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:1.2.3'
+        classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.12.+'
```
ecgreb: :+1:

lost-sample/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+    buildTypes {
+        //release {
+        //    minifyEnabled false
+        //    proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        //}
+    }
+}
+
+dependencies {
+    compile group: 'com.mapzen.android', name: 'lost', version:'0.4-SNAPSHOT'
```
ecgreb: Should be version 1.0.2-SNAPSHOT. Can we pull this from `gradle.properties` since it should always be the current version?

ecgreb:
Also we should delete all the pom.xml files. I don't really want to maintain two build systems do you? :stuck_out_tongue_winking_eye:

lost-sample/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+    }
+    buildTypes {
+        //release {
+        //    minifyEnabled false
+        //    proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        //}
+    }
+}
+
+dependencies {
```
ecgreb: Sample needs appcompat dependency to compile (uses AppCompat theme).

ecgreb:
Both `LostActivity` and `SettingsActivity` need now to extend `AppCompatActivity`. Also any calls to `getActionBar()` must be replaced with `getSupportActionBar()`.

crankycoder:
Ach. Sorry - been busy with regular day job stuff. I'll update the code as per your review tonight or tomorrow.

ecgreb:
@crankycoder no worries. I created a branch and pushed a commit on top of yours that allows me to build and run the sample app. See if it works for you as well. Also I have most of the unit tests working locally again with Robolectric 3.0-rc3. Will be pushing that soon.

https://github.com/mapzen/LOST/tree/chuck/gradlfication

ecgreb:
@crankycoder I'm going to merge this along with my additional commits to fix the sample app and unit tests. Let me know if you run into any issues when you start integrating MLS.

Issue #10, opened 2014-11-14
Support for Geofences?


maurizi:
Is there any intention to add support for Geofences and the Geofence API?

friesenkiwi:
I am currently trying to port OwnTracks from PlayServices to LOST (https://github.com/owntracks/android/issues/88). It looks good so far, but missing Geofencing would be a dealbreaker, since it is a feature many users of it heavily rely upon.
For the current development state, I added some try/catch for the "not yet implemented" RuntimeExceptions to at least not crash the app.
But we would still like to switch and thus "need" GeoFencing support in LOST. Are there any recent considerations of implementation or forseeable progress?
Thinking about it, it might be not too complicated to add it? As far as I can see, it would be implementing the three methods addGeofences(List<Geofence>, PendingIntent), removeGeofences(List<String>), removeGeofences(PendingIntent) in GeofencingApiImpl.java
and implement the Geofence.java interface, is that correct? Also, of course, the actual detection of transitions needs to be implemented. But since there are only circular Fences, that should be quite straightforward, right?
I found https://code.google.com/p/openmap/source/browse/src/openmap/com/bbn/openmap/geo/Intersection.java
or
http://www.codeproject.com/Articles/62482/A-Simple-Geo-Fencing-Using-Polygon-Method
as a start, but they seem to be eben a little bit too complex.
And I'd say, a minimal set of features would already be desireable, no need to implement all features at once? (Loitering, Dwelling, Responsiveness, GeofencingRequest etc.)

xbroak:
Hows this going? Any sign of addProximityAlert methods?

friesenkiwi:
Hey, there. Sorry I had to drop my first work on this last year, but I am right now looking into it again. I see @ecgreb you are working on it as well? How far is the progress? Would it be any helpful if I joined in?
Is the schedule for 2.0 very tight? Because I would love to see this feature in a release soon, so it can be used over at owntracks, and I would be very happy to help, if that's of interest.

ecgreb:
Hi @friesenkiwi. Its all good. Help is always appreciated. Here is what I have so far:

https://github.com/mapzen/LOST/pull/86

As you can see in the description things are not quite working yet. Also once inital support is in place there will be more work needed including registering multiple geofences, removing geofences, and triggering geofences via mock locations. Plus likely some other things I am overlooking.

friesenkiwi:
Hi! Awesome work, thank you for your code!
[...]

PR #31, opened 2015-08-05
Adding Geofencing  [WIP]

friesenkiwi:
Starting to tackle #10 
This is work in progress, I only implemented rough data structures, changed and added interfaces according to google docs, also there is one method deprecated now.

I just wanted to make sure, it looks OK for you and I didn't violate any code style or other guidelines?
Does it look OK to you so far? Is it worth the trouble to continue implementation or are you already on it?

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/api/Geofence.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/Geofence.java
@@ -1,42 +1,63 @@
 package com.mapzen.android.lost.api;
 
+import com.mapzen.android.lost.internal.GeofenceImpl;
+
 public interface Geofence {
 
-    public static final int GEOFENCE_TRANSITION_ENTER = 0x00000001;
-    public static final int GEOFENCE_TRANSITION_EXIT = 0x00000002;
-    public static final int GEOFENCE_TRANSITION_DWELL = 0x00000004;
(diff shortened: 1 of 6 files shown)


ecgreb:
This is great thanks for getting the geofencing work started.

I think we'll want to leverage classes in the framework as much as possible since the LocationManager already supports proximity alerts.

[https://developer.android.com/reference/android/location/LocationManager.html#addProximityAlert(double, double, float, long, android.app.PendingIntent)](https://developer.android.com/reference/android/location/LocationManager.html#addProximityAlert(double, double, float, long, android.app.PendingIntent))

As for code style for the most part this looks fine. I'm going to be adding checkstyle to the project shortly. Checks will be based on style defined at https://github.com/mapzen/open/blob/master/checkstyle.xml.

friesenkiwi:
OK, thank you for the encouragement and hints for checkstyle and proximity alerts! :-)
I will work on this to include the suggestions and missing features. Unfortunately, I will need to pause a bit on this for two or three weeks, so if anybody likes to work on it in the meantime, I'll be fine with that.

Is it OK for you if we first implemented a working and sensible subset of methods for a next release, or would you require a complete set as a whole?

ecgreb:
Working subset is fine. That's kind of what this project is all about ;)

ecgreb:
Closing due to inactivity. Feel free to re-open if you start up work again on this issue.

PR #41, opened 2015-12-18
Mock location bearing

### LINKED ISSUE #36: Populate bearing data in mock GPX replay locations
Can be calculated using previous and current coordinates.

http://developer.android.com/reference/android/location/Location.html#bearingTo(android.location.Location)

ecgreb:
Adds bearing to mock locations emitted by GPX replay. Uses previous and current coordinates to calculate approximate value.

Testing this feature enabled by Robolectric upgrade to fix thread scheduler problem that was causing mock locations to be delivered in reverse order when posted to main thread handler.

Updates sample app to display bearing value.

Closes #36

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/MockEngine.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/MockEngine.java
@@ -76,6 +76,7 @@ public class MockEngine extends LocationEngine {
 
     private class TraceThread extends Thread {
         private boolean canceled;
+        private Location previous;
 
         public void cancel() {
             canceled = true;
@@ -131,6 +132,11 @@ public class MockEngine extends LocationEngine {
             location.setTime(System.currentTimeMillis());
             location.setSpeed(Float.parseFloat(speedList.item(i).getFirstChild().getNodeValue()));
 
+            if (previous != null) {
+                location.setBearing(previous.bearingTo(location));
+            }
+
+            previous = location;
             return location;
         }
 
--- a/lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
+++ b/lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
@@ -8,7 +8,6 @@ import com.google.common.base.Charsets;
 import com.google.common.io.Files;
 
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.robolectric.RobolectricGradleTestRunner;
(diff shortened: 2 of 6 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/MockEngine.java
```diff
@@ -131,6 +132,11 @@ private Location nodeToLocation(NodeList nodeList, NodeList speedList, int i) {
             location.setTime(System.currentTimeMillis());
             location.setSpeed(Float.parseFloat(speedList.item(i).getFirstChild().getNodeValue()));
 
+            if (previous != null) {
+                location.setBearing(previous.bearingTo(location));
```
msmollin: So how do you get the bearing from startup?

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
```diff
@@ -252,7 +251,7 @@ public void setMockLocation_shouldInvokeListenerOnce() throws Exception {
         assertThat(listener.getMostRecentLocation()).isEqualTo(mockLocation);
     }
 
-    @Test @Ignore("Find a better way to test this without Thread.sleep()")
```
msmollin: I'm assuming we didn't find a better way? :)

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
```diff
@@ -260,7 +259,7 @@ public void setMockTrace_shouldInvokeListenerForEachLocation() throws Exception
         LocationRequest request = LocationRequest.create();
         request.setFastestInterval(0);
         api.requestLocationUpdates(request, listener);
-        Thread.sleep(1000);
+        Thread.sleep(100);
```
msmollin: Why the reduction here?

lost/src/main/java/com/mapzen/android/lost/internal/MockEngine.java
ecgreb: If there is no `previous` value the bearing is not set and `Location#hasBearing()` will return `false`.

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
ecgreb: Unfortunately no :(

However by upgrading Robolectric I was able to reduce the sleep value (see below) and ensure with greater certainty that things will run in the order we expect them to using controls like `ShadowLooper.runUiThreadTasks()`.

PR #46, opened 2016-02-19
Set default speed if not in gpx file

### LINKED ISSUE #45: Make .gpx speed list optional
Currently, `NodeList speedList` is mandatory or the app will crash while mocking the file. It should be a trivial modification to make it optional.

Most tracks available on openstreetmap lack speed tags, and the current workaround is to append a large number of `<speed>0</speed>` to the track segment.

Possibly related: it would be nice to make a gpx extension allowing accuracy, since this might be important while testing an app. Also see #37.

sarahlensing:
Closes #45


ecgreb:
Rather than setting a default speed I think it would be better to simply not set a value for speed if it does not exist in the trace. This is consistent with `Location` objects optionally not having a speed.

https://developer.android.com/reference/android/location/Location.html#hasSpeed()

This has also been discussed in https://github.com/mapzen/LOST/issues/45 and would resolve this issue.

sarahlensing:
@ecgreb That makes more sense than setting a default. I've updated it and added a test.

ecgreb:
:+1:

PR #51, opened 2016-05-27
Support multiple listeners in FusedLocationApi

### LINKED ISSUE #47: Manage multiple location listeners in FusedLocationProviderApi
Currently, only one LocationListener can be register at a given time in the implementation of the `FusedLocationProviderApi` (the [FusedLocationProviderApiImpl](https://github.com/mapzen/LOST/blob/master/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java)). It would be nice to have a list of listeners.

The reason: I currently develop an application with Mapbox, which use the `FusedLocationProviderApi`, but I also have another component that requires listening to location updates (using the LOST API). If my other component sends a request for location updates, the Mapbox component will not receive any more location updates because there is only one `LocationListener` in `FusedLocationProviderApiImpl` (and also because the `FusedLocationProviderApi` is static, which seems normal).

This modification would require a bit of coding to "combine" the [requests](https://github.com/mapzen/LOST/blob/master/lost/src/main/java/com/mapzen/android/lost/api/LocationRequest.java). For instance, if a request asks for a `PRIORITY_HIGH_ACCURACY` and another for `PRIORITY_LOW_POWER`, then the `PRIORITY_HIGH_ACCURACY` will be the effective priority. Same for the updates intervals and displacement.

sarahlensing:
* Creates new `LocationEngine` per `LocationRequest`
* Adds/removes listeners for given `LocationEngine`
* Shuts `LocationEngine` down when no listeners exist for it
* Requires removal of all listeners before calling `LostApiClient#disconnect()` to properly shutdown client

Closes #47

### DIFF
--- a/lost-sample/build.gradle
+++ b/lost-sample/build.gradle
@@ -55,8 +55,8 @@ task checkstyle(type: Checkstyle) {
 }
 
 dependencies {
-  compile group: 'com.mapzen.android', name: 'lost', version: '1.0.2-SNAPSHOT'
-  compile group: 'com.mapzen.android', name: 'prefs-plus', version: '1.0.0'
+  compile project(':lost')
+  compile 'com.mapzen.android:prefs-plus:1.0.0'
   compile 'com.android.support:appcompat-v7:22.2.0'
 }
 
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -10,6 +10,9 @@ import android.location.Location;
 import android.os.Looper;
 
 import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 
 /**
  * Implementation of the {@link FusedLocationProviderApi}.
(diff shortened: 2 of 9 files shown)


lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
```diff
@@ -65,7 +65,7 @@ public void disconnect_shouldRemoveGeofencingApiImpl() throws Exception {
     }
 
     @Test
-    public void disconnect_shouldUnregisterLocationUpdateListeners() throws Exception {
+    public void disconnect_shouldNotUnregisterLocationUpdateListeners() throws Exception {
```
ecgreb: Are we sure this is the behavior we want? This seems like it could lead to unintentionally leaving location services running in the background.

I wonder if it would be better to remove _all_ listeners (and shutdown all engines) when disconnect is called.

sarahlensing: I had thought it would be good to enforce best practices on the user of this api but there isn't any harm in removing all listeners and shutting down as you suggest. I'll update the code.

PR #54, opened 2016-06-20
Dont remove listeners on setMockMode

sarahlensing:
- Instead of shutting down all engines when mock mode is toggled on/off, toggle them so that `MockEngine`s are created for existing listeners when mock mode enabled
- Add method to get the number of listeners so that shared instances of `LostApiClient` can know whether or not they should call `disconnect`

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -80,7 +80,7 @@ public class FusedLocationProviderApiImpl implements
 
     private void toggleMockMode() {
         mockMode = !mockMode;
-        shutdownAllEngines();
+        toggleAllEngines();
         if (mockMode) {
             lastLocationEngine = new MockEngine(context, null);
         } else {
@@ -121,6 +121,14 @@ public class FusedLocationProviderApiImpl implements
 
     public void shutdown() {
         shutdownAllEngines();
+    }
+
+    public List<LocationListener> getListeners() {
+        List<LocationListener> listeners = new ArrayList<>();
+        for (LocationEngine engine : engineListeners.keySet()) {
+            listeners.addAll(engineListeners.get(engine));
+        }
+        return listeners;
     }
 
     /**
@@ -180,6 +188,39 @@ public class FusedLocationProviderApiImpl implements
             }
         }
         return null;
+    }
+
+    private void toggleAllEngines() {
+        HashMap<LocationRequest, LocationEngine> enginesToAdd = new HashMap<>();
+        HashMap<LocationEngine, List<LocationListener>> listenersToAdd = new HashMap<>();
+        for (LocationRequest request : locationEngines.keySet()) {
+            LocationEngine existing = locationEngines.get(request);
+            LocationEngine engineToAdd;
+            if (mockMode) {
+                engineToAdd = new MockEngine(context, this);
+            } else {
+                engineToAdd = new FusionEngine(context, this);
+            }
+            enginesToAdd.put(request, engineToAdd);
+
+            List<LocationListener> listeners = engineListeners.get(existing);
+            listenersToAdd.put(engineToAdd, listeners);
+
+            //Cleanup listeners
+            engineListeners.remove(existing);
+            existing.setRequest(null);
+        }
+        //Cleanup engines
+        locationEngines.clear();
+
+        //Now add new engines/listeners
+        for (LocationRequest request : enginesToAdd.keySet()) {
+            LocationEngine engine = enginesToAdd.get(request);
+            List<LocationListener> listeners = listenersToAdd.get(engine);
+            for (LocationListener listener : listeners) {
+                requestLocationUpdates(request, listener);
+            }
+        }
     }
 
     private void shutdownAllEngines() {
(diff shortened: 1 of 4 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -123,6 +123,14 @@ public void shutdown() {
         shutdownAllEngines();
     }
 
+    public List<LocationListener> getListeners() {
```
ecgreb: Since we decided to remove `LostApiClient#numberOfListeners()` can this also be removed?

I see its currently used in tests only. Can we switch back to using the `ShadowLocationManager` in those tests?

sarahlensing: I think we still need this because as far as I can tell, `assertThat(shadowLocationManager.getRequestLocationUpdateListeners()).isEmpty();` will always be true when mock mode is set. Maybe we could set this visible for testing only?

ecgreb: Ok now I understand the confusion. These tests were meant to test listeners registered by LOST on the framework `LocationManager` (not by client apps on LOST). See comments below.

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
```diff
@@ -203,12 +203,12 @@ public void removeLocationUpdates_shouldUnregisterAllListeners() throws Exceptio
     }
 
     @Test
-    public void setMockMode_shouldUnregisterAllListenersWhenTrue() throws Exception {
+    public void setMockMode_shouldToggleAllListenersWhenTrue() throws Exception {
```
ecgreb: The original purpose of this test was to ensure all location listeners were removed from the `LocationManager` so we are not invoking system location services while in mock mode. I think that test is still valid in its original form but maybe we want to update the name to something like `setMockMode_shouldRemoveAllLocationManagerListeners()`.

Perhaps we should add an additional test to verify the correct number of listeners has been set on the `FusedLocationProviderApi` by the client app.

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
```diff
@@ -219,7 +219,7 @@ public void setMockMode_shouldNotRegisterDuplicateListeners() throws Exception {
         api.requestLocationUpdates(request, listener);
         api.setMockMode(false);
         api.requestLocationUpdates(request, listener);
-        assertThat(shadowLocationManager.getRequestLocationUpdateListeners()).hasSize(2);
+        assertThat(api.getListeners()).hasSize(2);
```
ecgreb: Similarly this test was added to ensure we were not registering duplicate listeners with the `LocationManager` when exiting mock mode (which was an actual bug at some point)

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -182,11 +183,55 @@ private LocationEngine removeListenerFromEngine(LocationListener listener) {
...
     private void shutdownAllEngines() {
         for (LocationEngine engine : locationEngines.values()) {
             engineListeners.remove(engine);
             engine.setRequest(null);
         }
         locationEngines.clear();
     }
+
+    @VisibleForTesting
+    public List<LocationListener> getListeners() {
```
ecgreb: Package private?

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
```diff
@@ -219,7 +219,17 @@ public void setMockMode_shouldNotRegisterDuplicateListeners() throws Exception {
         api.requestLocationUpdates(request, listener);
         api.setMockMode(false);
         api.requestLocationUpdates(request, listener);
-        assertThat(shadowLocationManager.getRequestLocationUpdateListeners()).hasSize(2);
+        assertThat(shadowLocationManager.getRequestLocationUpdateListeners()).hasSize(4);
```
ecgreb: It seems like `shadowLocationManager.getRequestLocationUpdateListeners()` should still return `2` if the original listeners were properly removed. There should be one listener for the GPS provider and one for the network provider.

I stepped through this test with the debugger. What seems to be happening is the first time `requestLocationUpdates(...)` is called no listeners are registered b/c mock mode is true. When mock mode is set to `false` then `toggleAllEngines()` registers two listeners (GPS + network) with the `LocationManager`. Then when `requestLocationUpdates(...)` is called for the second time the same two listeners are registered with the `LocationManager` yet again.

I'm not sure what the implications are of registering the same listener twice with the `LocationManager`. Perhaps it doesn't matter. Or perhaps we can set a flag to when the engine has been registered with the `LocationManager` to ensure the same engine does not get registered more than once.

sarahlensing: True, we should only register an engine if it hasn't be registered before. I've updated the code to check if the engine is existing and if not, register it!

ecgreb:
👍

PR #56, opened 2016-07-18
Adds isProviderEnabled & onProviderDisabled/Enabled callbacks

### LINKED ISSUE #55: Q: isProviderEnabled & onProviderDisabled/Enabled callbacks
Here is a little question, I know this is not the way how the fused location provider works, but is there a way to receive a status of enabled providers and get appropriate callbacks or I should request android.location directly?

ecgreb:
**Library**
* Exposes ability to query if a location provider is enabled.
* Adds location listener callbacks for provider enabled/disabled.

**Sample**
* Adds logging for provider enabled/disabled callbacks.
* Adds setting to enable location updates when sample app is running in the background.

Closes #55


sarahlensing:
👍

PR #60, opened 2016-07-22
Enforce 1-1 request/listener relationship

### LINKED ISSUE #57: Enforce one-to-one relationship for requests and listeners
Each `LocationRequest` should have only one `LocationListener` and each `LocationListener` should be associated with only one `LocationRequest`.

This also implies that each request/listener pair would have its own `FusionEngine` using the current architecture and that engine would not be re-used for other requests or listeners.

Alternatively we could adopt a model that uses a single `FusionEngine` for all requests/listeners. This is how it's [performed in the Android framework](https://github.com/android/platform_frameworks_base/blob/master/packages/FusedLocation/src/com/android/location/fused/FusionEngine.java) and would require implementing some form of [request bundling.](https://github.com/android/platform_frameworks_base/blob/master/location/lib/java/com/android/location/provider/ProviderRequestUnbundled.java)

Either way we decide to implement this it should simplify the logic required in `FusedLocationProviderApiImpl` (especially when toggling mock mode) and better match the behavior of the [play services APIs](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderApi#requestLocationUpdates(com.google.android.gms.common.api.GoogleApiClient, com.google.android.gms.location.LocationRequest, com.google.android.gms.location.LocationListener)).

sarahlensing:
- Use single `LocationEngine` in `FusedLocationProviderApiImpl`
- Bundle `LocationRequest` into `LocationRequestUnbundled`

As a result of these changes, listeners may receive results that are more/less accurate than requested but I think this is okay given that the LocationRequest documentation says this:
"All location requests are considered hints, and you may receive locations that are more/less accurate, and faster/slower than requested."

https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest

Closes #57

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -10,9 +10,8 @@ import android.location.Location;
 import android.os.Looper;
 
 import java.io.File;
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
+import java.util.Map;
 
 /**
  * Implementation of the {@link FusedLocationProviderApi}.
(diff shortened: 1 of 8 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -84,182 +74,61 @@ public void setMockMode(boolean isMockMode) {
...
-                }
-                return engine;
-            }
-        }
-        return null;
+    public Map<LocationListener, LocationRequest> getListeners() {
+        return listenerToRequest;
     }
 
-    private void toggleAllEngines() {
```
ecgreb: This refactor was a huge win in simplifying this logic. Great.

lost/src/main/java/com/mapzen/android/lost/internal/FusionEngine.java
```diff
@@ -99,33 +101,33 @@ protected void disable() {
         }
     }
 
-    private void enableGps() {
+    private void enableGps(LocationRequest request) {
         try {
             locationManager.requestLocationUpdates(GPS_PROVIDER,
-                    getRequest().getFastestInterval(),
-                    getRequest().getSmallestDisplacement(),
+                    request.getFastestInterval(),
```
ecgreb: Currently all listeners will receive updates according the fastest interval and smallest displacement params from the last request in the list. While I think its fine for listeners to receive updates more frequently than requested it seems wrong to send updates less frequently because there was a more recent request with a slower interval.

Instead we should loop over all the requests and sent the lowest value among all requests for the interval. In other words we should use the _fastest_ fastest interval. In fact this is how its handled in the Android framework for setting the interval on an individual provider.

https://github.com/android/platform_frameworks_base/blob/4535e11fb7010f2b104d3f8b3954407b9f330e0f/packages/FusedLocation/src/com/android/location/fused/FusionEngine.java#L156-L197

The other interesting thing is that in the framework the minimum displacement used by the `FusionEngine` is always `0` regardless of the actual value in the location requests. I wonder if we should be doing the same thing?

https://github.com/android/platform_frameworks_base/blob/4535e11fb7010f2b104d3f8b3954407b9f330e0f/packages/FusedLocation/src/com/android/location/fused/FusionEngine.java#L129-L144

PR #72, opened 2016-08-11
Request location updates via pending intents

### LINKED ISSUE #22: Location updates via PendingIntent

ecgreb:
* Enable providers for requests via PendingIntent
* Report location updates via PendingIntent

~~TODO: Remove pending intent updates~~ 97cae218dd16d99dc7f1b3fa0536a3179a78fab8

Closes #22


ecgreb:
I added support for the "new" way of attaching location updates to a `PendingIntent` using the `LocationResult` class. This commit also deprecates `FusedLocationProviderApi.KEY_LOCATION_CHANGED`.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -62,7 +69,10 @@ public void removeLocationUpdates(LocationListener listener) {
 
     @Override
     public void removeLocationUpdates(PendingIntent callbackIntent) {
-        throw new RuntimeException("Sorry, not yet implemented");
+        intentToRequest.remove(callbackIntent);
+        if (intentToRequest.isEmpty()) {
```
sarahlensing: Should we check that `listenerToRequest` is null before setting the engine's request to null? And should we check that `intentToRequest` is null in L65? It feels slightly wrong to stop receiving location updates in the engine if there are still listeners but no pending requests (and vice versa).

ecgreb: Good catch. Although probably we want to check if `listenerToRequest` and `intentToRequest` are both empty rather than null, right? Updated in 8c0c5c1276f1fde195b71e543e64aa2a1218b794.

bleege:
👍

PR #74, opened 2016-08-12
Add SettingsApi for resolving location status

sarahlensing:
- Adds `SettingsApi` for requesting location and ble status/updating status
- Adds sample activity and restructures the sample app to show examples categorized by api

A user of this api can now construct a `LocationSettingsRequest` and use `LocationServices.SettingsApi#checkLocationSettings` to receive a `LocationSettingsResult`. The user can then inspect `LocationSettingsStates` and `Status` information to know which location settings are enabled & usable. If a resolution of state is needed, `Status#startResolutionForResult` can be invoked. This will display the device's Settings activity for location management. To then update a client application's state, the user should implement `Activity#onActivityResult` and check only that the `resultCode` is what is expected before refreshing UI or state. 

~~TODO: implement remaining empty tests~~ ~~implement 4 remaining tests that deal with AsyncTask~~ ca2cdea551fbb724bc175359d623502b8a2840d7

Closes #74


lost/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,14 @@
         android:required="false" />
     <uses-feature android:name="android.hardware.wifi" />
 
-    <!-- Wow. This is insane. You require an empty tag. -->
-    <application />
+
+    <uses-permission android:name="android.permission.BLUETOOTH"/>
```
ecgreb: I'm concerned about including bluetooth permissions in the library manifest. Developers using lost are typically sensitive (and rightfully so) about adding unnecessary permissions to their apps.

Would it be possible to exclude these permissions from the library manifest? Then we could specify in the README if you want to access bluetooth settings using lost you must add these permissions to the app manifest.

lost/build.gradle
```diff
@@ -69,6 +69,7 @@ dependencies {
   testCompile 'org.robolectric:robolectric:3.1.2'
   testCompile 'com.squareup:fest-android:1.0.7'
   testCompile 'org.mockito:mockito-core:1.9.5'
+  compile 'com.android.support:support-annotations:23.4.0'
```
ecgreb: I added these in https://github.com/mapzen/LOST/pull/72 as well ;) Very handy.

lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
```diff
@@ -0,0 +1,218 @@
...
+    boolean needBle = settingsRequest.getNeedBle();
+
+
+    boolean gpsUsable = locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER);
+    boolean gpsPresent = packageManager.hasSystemFeature(PackageManager.FEATURE_LOCATION_GPS);
+    boolean networkUsable = locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER);
+    boolean networkPresent = packageManager.hasSystemFeature(
+        PackageManager.FEATURE_LOCATION_NETWORK);
+    boolean bleUsable = bluetoothAdapter.isEnabled();
+    boolean blePresent = packageManager.hasSystemFeature(PackageManager.FEATURE_BLUETOOTH_LE);
```
ecgreb: `PackageManager.FEATURE_BLUETOOTH_LE` requires API 18. The current minimum SDK for LOST is 15.

lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
```diff
@@ -0,0 +1,218 @@
...
+
+import static com.mapzen.android.lost.api.Status.CANCELLED;
+import static com.mapzen.android.lost.api.Status.INTERNAL_ERROR;
+import static com.mapzen.android.lost.api.Status.INTERRUPTED;
+import static com.mapzen.android.lost.api.Status.RESOLUTION_REQUIRED;
+import static com.mapzen.android.lost.api.Status.SETTINGS_CHANGE_UNAVAILABLE;
+import static com.mapzen.android.lost.api.Status.SUCCESS;
+import static com.mapzen.android.lost.api.Status.TIMEOUT;
+
+public class LocationSettingsResultRequest extends PendingResult<LocationSettingsResult> {
```
ecgreb: After reading and testing the code this class certainly seems like it works as intended. Nicely done.

However I'm still hesitant to introduce this level of concurrency into the the framework for what are essentially short-lived operations (`isProviderEnabled()`, `hasSystemFeature()`, etc) that could be executed on the UI thread with no real drawbacks.

Here are some questions that come to mind:

1. What benefits do we get from wrapping the `FutureTask` in an `AsyncTask`? Could we just use a `FutureTask` directly?
2. Is it possible to implement a version of the `PendingResult` API without the actual use of concurrency? Can we "fake" it until we actually have a need for it?
3. Should we eliminate the `PendingResult` API all together? In our quest to achieve API parity with play services are we introducing code/features that don't really make sense in the context of the LOST framework?

Play services has a ton of overhead due to authentication and IPC that we don't have to deal with (at least not yet) which probably makes more sense for asynchronous operations.

Concurrency can introduce subtle bugs in the code that can be hard to track down. It also makes unit testing more challenging. So before we commit to this path forward I just want to make sure we are considering all our options and the pros/cons of each.

lost/src/main/AndroidManifest.xml
sarahlensing: Good point. I've removed them from the library and updated our README to document their requirement when using this API

lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
sarahlensing: Nice catch!

So it looks like we wont crash when making this call on devices < API 18 https://android.googlesource.com/platform/frameworks/base/+/483f3b06ea84440a082e21b68ec2c2e54046f5a6/services/java/com/android/server/pm/PackageManagerService.java#1861

But after testing on an emulator running API 17, the `BluetoothAdapter` is null so I've added a check for this in L181. I think with this in place, its safe for us to keep the minimum where it is and achieve expected behavior on devices running lower APIs.

lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
sarahlensing: Thanks for the thoughtful feedback. Lets address these questions:

1. Upon more thought, we don't get a real benefit from wrapping the `FutureTask` since the operations are short-lived as you say. It was implemented that way with future complication in mind so we definitely could (and probably should, now that you mention it) simplify it and could use a `FutureTask` directly.
2. We could fake the concurrency and implement something similar to what we do in `await()` but just return the `LocationSettingsResult` via the `ResultCallback` instead of having the method return it.
3. This is an interesting question. I think it might be a drastic move though and break too much with the current parity we have across the rest of the library. I think that faking the concurrency should be enough to balance keeping our library simple and users familiar with the api. I'd vote to try out this simplification before eliminating the API.

How does keeping the api and faking the concurrency sound as a path forward?

ecgreb: The seems like the most balanced and best approach for now. Let's do it.

lost/src/main/java/com/mapzen/android/lost/internal/PendingIntentGenerator.java
```diff
@@ -0,0 +1,25 @@
+package com.mapzen.android.lost.internal;
+
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.provider.Settings;
+
+public class PendingIntentGenerator {
+
+  Context context;
```
ecgreb: private final?

lost-sample/src/main/java/com/example/lost/SettingsApiActivity.java
```diff
@@ -0,0 +1,205 @@
...
+    boolean needBle = true;
+    LocationSettingsRequest request = new LocationSettingsRequest.Builder()
+        .addAllLocationRequests(requests)
+        .setNeedBle(needBle)
+        .build();
+    LostApiClient apiClient = new LostApiClient.Builder(this).build();
+    apiClient.connect();
+    result = LocationServices.SettingsApi.checkLocationSettings(apiClient, request);
+
+    //result.setResultCallback(new ResultCallback<LocationSettingsResult>() {
```
ecgreb: I get why this is here to show the different ways to invoke the settings API but I don't think we want to commit so much commented out code to the sample app as it makes it harder to follow.

If you think this is helpful to keep in the example what about refactoring this code into a private method then then just commenting out the method call?

lost/build.gradle
ecgreb: We should manually rebase these changes onto master before merging otherwise the support-annotations library will be included twice when the auto-merge resolves.

https://github.com/mapzen/LOST/blob/master/lost/build.gradle#L67

lost/src/test/java/com/mapzen/android/lost/internal/LocationSettingsResultRequestTest.java
```diff
@@ -0,0 +1,606 @@
...
+        .addAllLocationRequests(requests)
+        .setNeedBle(true)
+        .build();
+
+    resultRequest = new LocationSettingsResultRequest(bluetoothAdapter, pm, locationManager,
+        generator, request);
+  }
+
+  @Test
+  public void await_shouldReturnSuccessfullResult() {
```
ecgreb: Typo

lost/src/test/java/com/mapzen/android/lost/internal/LocationSettingsResultRequestTest.java
```diff
@@ -0,0 +1,606 @@
...
+    DelayTestPendingIntentGenerator delayedGenerator = new DelayTestPendingIntentGenerator(
+     context);
+    final LocationSettingsResultRequest settingsResultRequest = new LocationSettingsResultRequest(
+     bluetoothAdapter, pm, locationManager, delayedGenerator, request);
+
+    Thread thread = new Thread(new Runnable() {
+      @Override public void run() {
+        settingsResultRequest.setResultCallback(new ResultCallback<LocationSettingsResult>() {
+          @Override public void onResult(@NonNull LocationSettingsResult result) {
+            assertThat(result.getStatus().getStatusCode()).isEqualTo(Status.CANCELLED);
```
ecgreb: The problem with putting assert statements in a background thread is that even if the assertion fails the test will still pass. You can replace the asserts here with `assertThat(false).isTrue()` and the test will still not fail.

lost/src/test/java/com/mapzen/android/lost/internal/LocationSettingsResultRequestTest.java
```diff
@@ -0,0 +1,606 @@
...
+        generator, request);
+  }
+
+  @Test
+  public void await_shouldReturnSuccessfullResult() {
+    when(locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER)).thenReturn(true);
+    when(pm.hasSystemFeature(PackageManager.FEATURE_LOCATION_GPS)).thenReturn(true);
+    when(locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER)).thenReturn(true);
+    when(pm.hasSystemFeature(PackageManager.FEATURE_LOCATION_NETWORK)).thenReturn(true);
+    when(bluetoothAdapter.isEnabled()).thenReturn(true);
```
ecgreb: We should add `@SuppressWarnings("MissingPermission") ` to this class so Android Studio inspections stop complaining about permissions in tests.

lost/src/main/java/com/mapzen/android/lost/internal/ResolveLocationActivity.java
```diff
@@ -0,0 +1,33 @@
...
+ * Used to enable bluetooth before firing intent to open location settings activity.
+ */
+public class ResolveLocationActivity extends Activity {
+
+  private static final int REQUEST_CODE = 1;
+
+  @Override protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+
+    BluetoothAdapter adapter = BluetoothAdapter.getDefaultAdapter();
```
ecgreb: Per documentation `BluetoothAdapter.getDefaultAdapter()` can return null "if Bluetooth is not supported on this hardware platform". We should check for null before calling `adapter.enable()`.

https://developer.android.com/reference/android/bluetooth/BluetoothAdapter.html#getDefaultAdapter()

lost/src/main/java/com/mapzen/android/lost/internal/PendingIntentGenerator.java
sarahlensing: Updated

lost-sample/src/main/java/com/example/lost/SettingsApiActivity.java
sarahlensing: Good point, commented out code has been removed.

lost/src/test/java/com/mapzen/android/lost/internal/LocationSettingsResultRequestTest.java
sarahlensing: Fixed

lost/src/test/java/com/mapzen/android/lost/internal/LocationSettingsResultRequestTest.java
sarahlensing: I don't have a good way to cleanly test this at the moment so I've removed the test. Let me know if you see a better way forward though.

lost/src/test/java/com/mapzen/android/lost/internal/LocationSettingsResultRequestTest.java
sarahlensing: Added

lost/src/main/java/com/mapzen/android/lost/internal/ResolveLocationActivity.java
sarahlensing: Nice catch- added

lost/build.gradle
ecgreb: This should be removed as it already exists on L67.

sarahlensing: Rm'd

PR #81, opened 2016-08-17
Remove location permissions

### LINKED ISSUE #75: Remove location permissions from library
The LOST library manifest should not include permissions `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION`. It should be the responsibility of client applications to include necessary permissions in its manifest based on usage.

Requires permission annotations should be added to the appropriate methods in the `FusedLocationProviderApi`.

```java
    @RequiresPermission(
        anyOf = {"android.permission.ACCESS_COARSE_LOCATION", "android.permission.ACCESS_FINE_LOCATION"}
    )
```

ecgreb:
### Overview

Permissions `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION` should not be included in the LOST library `AndroidManifest.xml`. Instead they should be added as required by client applications based on their API use and additionally requested at runtime on API 23+.

### Proposed Changes

Location permissions have been removed from the library manifest and added to the sample application. `FusedLocationProviderApi` has been modified to use support annotations for `@RequiresPermission`. Finally `throws SecurityException` has been explicitly added to all `FusionEngine` methods that may be rejected if permissions have not been granted by the user.

Closes #75

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/api/FusedLocationProviderApi.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/FusedLocationProviderApi.java
@@ -1,7 +1,6 @@
 package com.mapzen.android.lost.api;
 
 import android.app.PendingIntent;
-import android.content.Intent;
 import android.location.Location;
 import android.os.Looper;
 import android.support.annotation.RequiresPermission;
@@ -14,8 +13,8 @@ import static android.Manifest.permission.ACCESS_FINE_LOCATION;
 public interface FusedLocationProviderApi {
 
   /**
-   * @deprecated Use {@link LocationResult#hasResult(Intent)} and
-   * {@link LocationResult#extractResult(Intent)}.
+   * @deprecated Use {@link LocationResult#hasResult(android.content.Intent)} and
+   * {@link LocationResult#extractResult(android.content.Intent)}.
    */
   @Deprecated String KEY_LOCATION_CHANGED = "com.mapzen.android.lost.LOCATION";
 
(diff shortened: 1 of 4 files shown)


lost/src/main/java/com/mapzen/android/lost/api/FusedLocationProviderApi.java
```diff
@@ -3,30 +3,49 @@
...
+  @RequiresPermission(anyOf = {ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION})
   void requestLocationUpdates(LocationRequest request, PendingIntent callbackIntent);
 
+  @RequiresPermission(anyOf = {ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION})
   void setMockLocation(Location mockLocation);
 
+  @RequiresPermission(anyOf = {ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION})
   void setMockMode(boolean isMockMode);
 
+  @RequiresPermission(anyOf = {ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION})
```
sarahlensing: It looks like we technically dont need to require location permissions for the three setMock* methods. Could we remove the annotation on them?

ecgreb: Good point. I originally added them since they exist in the play services version. But since our version of mocking does not use the `LocationManager` these can be safely dropped. Removed.

I also moved around the methods in `FusedLocationProviderApi` to follow a more logical order.

PR #83, opened 2016-08-18
63 background service

### LINKED ISSUE #63: Background service
Explore the option of running the fused location provider in a background service.

sarahlensing:
### Overview
This PR creates a service for the `FusedLocationProviderApi` to run in. When clients connect, the service will be started if it isn't already running. When clients disconnect, the service will ~~remain running (until the system kills it)~~ be shutdown.

### Proposed Changes
These changes move the `FusedLocationProviderApiImpl` code into a class called `FusedLocationProviderServiceImpl` which is instantiated and invoked inside of the new `FusedLocationProviderService`. With this change, we now have a new interface, `ConnectionCallbacks`, which client applications should use to know when they can start interacting with the `FusedLocationApi`, etc APIs.

Closes #63

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -3,46 +3,69 @@ package com.mapzen.android.lost.internal;
 import com.mapzen.android.lost.api.FusedLocationProviderApi;
 import com.mapzen.android.lost.api.LocationListener;
 import com.mapzen.android.lost.api.LocationRequest;
+import com.mapzen.android.lost.api.LostApiClient;
 
 import android.app.PendingIntent;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
+import android.content.ServiceConnection;
 import android.location.Location;
+import android.os.IBinder;
 import android.os.Looper;
-import android.util.Log;
 
 import java.io.File;
-import java.util.HashMap;
 import java.util.Map;
 
 /**
  * Implementation of the {@link FusedLocationProviderApi}.
  */
 public class FusedLocationProviderApiImpl
-    implements FusedLocationProviderApi, LocationEngine.Callback {
-
-  private static final String TAG = FusedLocationProviderApiImpl.class.getSimpleName();
+    implements FusedLocationProviderApi {
 
   private final Context context;
-  private boolean mockMode;
-  private LocationEngine locationEngine;
-  private Map<LocationListener, LocationRequest> listenerToRequest;
-  private Map<PendingIntent, LocationRequest> intentToRequest;
+  private FusedLocationProviderService service;
+
+  private final ServiceConnection serviceConnection = new ServiceConnection() {
+    @Override public void onServiceConnected(ComponentName name, IBinder binder) {
+      FusedLocationProviderService.FusedLocationProviderBinder fusedBinder =
+          (FusedLocationProviderService.FusedLocationProviderBinder) binder;
+      service = fusedBinder.getService();
+
+      if (connectionCallbacks != null) {
+        connectionCallbacks.onConnected();
+      }
+    }
+
+    @Override public void onServiceDisconnected(ComponentName name) {
+      if (connectionCallbacks != null) {
+        connectionCallbacks.onConnectionSuspended();
+      }
+    }
+  };
+
+  LostApiClient.ConnectionCallbacks connectionCallbacks;
 
   public FusedLocationProviderApiImpl(Context context) {
     this.context = context;
-    locationEngine = new FusionEngine(context, this);
-    listenerToRequest = new HashMap<>();
-    intentToRequest = new HashMap<>();
+  }
+
+  public void connect(LostApiClient.ConnectionCallbacks callbacks) {
+    connectionCallbacks = callbacks;
+    Intent intent = new Intent(context, FusedLocationProviderService.class);
+    context.bindService(intent, serviceConnection, Context.BIND_AUTO_CREATE);
+  }
+
+  public void disconnect() {
+    context.unbindService(serviceConnection);
   }
 
   @Override public Location getLastLocation() {
-    return locationEngine.getLastLocation();
+    return service.getLastLocation();
   }
 
   @Override public void requestLocationUpdates(LocationRequest request, LocationListener listener) {
-    listenerToRequest.put(listener, request);
-    locationEngine.setRequest(request);
+    service.requestLocationUpdates(request, listener);
   }
 
   @Override public void requestLocationUpdates(LocationRequest request, LocationListener listener,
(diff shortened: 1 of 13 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -3,46 +3,79 @@
...
+    Intent intent = new Intent(context, FusedLocationProviderService.class);
+    context.startService(intent);
+
+    connectionCallbacks = callbacks;
+    intent = new Intent(context, FusedLocationProviderService.class);
+    context.bindService(intent, serviceConnection, Context.BIND_AUTO_CREATE);
+  }
+
+  public void disconnect() {
+    context.unbindService(serviceConnection);
```
ecgreb: We should probably also call `context.stopService()` after the the client is no longer bound so it does not continue to consume resources when no clients are connected.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public FusedLocationProviderServiceImpl(Context context) {
+    this.context = context;
+    locationEngine = new FusionEngine(context, this);
+    listenerToRequest = new HashMap<>();
+    intentToRequest = new HashMap<>();
+  }
+
+  public void shutdown() {
+    listenerToRequest.clear();
```
ecgreb: I missed this when adding support for `PendingIntent` updates but `shutdown()` should also clear `intentToRequest`.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
sarahlensing: If we do this, we should probably not explicitly start the service with `context.startService(intent)` and let it shutdown automatically when all clients are unbound from it. How does this sound?

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
sarahlensing: Good catch

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
ecgreb: That sounds like it should be fine but we should test the following use case:

1. A location client is created and connected by an Activity.
2. The activity reqeusts location updates via PendingIntent where the receiver of the updates is an IntentService.
3. The activity is destroyed without calling disconnect on the Lost client.
4. The FusedLocationProviderService should continue to run sending location updates to the IntentService.

sarahlensing: This use case will succeed in sending location updates to the IntentService as long as the Context provided to the LostApiClient is not the Activity's. So if the application context is used: `LostApiClient.Builder(getApplicationContext()).addConnectionCallbacks(this).build();` this use case works. I don't think this is what you intend though...but I don't know if there is a way around this using the bound service approach.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
sarahlensing: Fixed

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
sarahlensing: ...continuing from this last thought...If we keep the explicit starting of the service in place, this use case will work. So maybe we SHOULD keep the explicit starting of the service and add a call to stop the service as you suggest. I was originally thinking that this could lead to unexpected behavior if there were multiple clients because if one client called `disconnect()` it would stop the service that others were bound to but I realize now that there will only ever be one client bound to the service because of L27 in LostApiClientImpl.

So lets do as you say and add a call to explicitly stop the service... 👍

ecgreb: Ha ok I didn't think about using the application context. This seems fine if not the preferred behavior. It makes sense that if you use an activity as the context the lifecycle of the connection would be bound to the lifecycle of the activity and if you use the application context it would not. Is this how play services behaves too?

ecgreb: So apparently play services does not agree with me. Location updates sent to an `IntentService` continue even when using an activity context to connect even after the activity is destroyed so we should probably do the same. I guess that means explicit calls to start/stop is the way to go!

lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
```diff
@@ -67,26 +61,8 @@
...
-          }
-        });
-
-    client.disconnect();
-    LocationManager lm = (LocationManager) application.getSystemService(LOCATION_SERVICE);
-    assertThat(shadowOf(lm).getRequestLocationUpdateListeners()).isEmpty();
-  }
-
-  @Test public void isConnected_shouldReturnFalseBeforeConnected() throws Exception {
+  @Test @Ignore("Failing after removal of disconnect_shouldUnregisterLocationUpdateListeners")
```
ecgreb: This test is not failing for me.

It seems like it should be fine since the `tearDown()` method is calling `client.disconnect()` after each test method.

lost/src/test/java/com/mapzen/android/lost/internal/MockEngineTest.java
```diff
@@ -124,7 +124,8 @@ public void setTrace_shouldNotRequireSpeed() throws Exception {
     assertThat(callback.locations.get(0).hasSpeed()).isFalse();
   }
 
-  @Test public void disable_shouldCancelTraceReplay() throws Exception {
+  @Test @Ignore("Failing")
```
ecgreb: Seems related to all the other tests currently ignored with the comment `"Intermittently failing. Find a better way to test without Thread.sleep(100)"`. Let's use the same comment here and I'll add an issue to deal with all of these at some point.

ecgreb: https://github.com/mapzen/LOST/issues/84

lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
sarahlensing: It doesn't fail for me when I run it by itself but it looks like it also fails on circle when run with the rest of the tests: https://circleci.com/gh/mapzen/LOST/242

I added code to set the `LocationServices` apis to null. Maybe there is a better way to resolve this though?

ecgreb: At first I was concerned that calling `client.disconnect()` wasn't working the way it should. But it turns out it was. But because `isConnected_shouldReturnFalseBeforeConnected` was the first test running in the class `LostApiClientImplTest` on Circle CI `tearDown()` was not run yet and the client was still connected from a previous test class.

9ac9e10643718cf3952202352424b05d415baff6 seems to have resolved the issue.

PR #85, opened 2016-08-23
64 location availability

### LINKED ISSUE #64: Implement getLocationAvailability and onLocationAvailability
Currently LOST provides no way to determine if a devices location services are enabled.

The Google Play API provides [LocationServices.FusedLocationApi.getLocationAvailability()](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderApi#getLocationAvailability(com.google.android.gms.common.api.GoogleApiClient)) and [LocationCallback.onLocationAvailability()](https://developers.google.com/android/reference/com/google/android/gms/location/LocationCallback#onLocationAvailability(com.google.android.gms.location.LocationAvailability)).

sarahlensing:
### Overview
This PR adds the ability to query for `LocationAvailability` as well as the ability to request location updates for a given `LocationCallback`. It also makes `PendingIntent`s more complete with information and adds extras `EXTRA_LOCATION_AVAILABILITY` and `EXTRA_LOCATION_RESULT` to be delivered with each `Intent`.

### Proposed Changes
The `FusedLocationProviderApi` api has three new methods (`getLocationAvailability()`, `requestLocationUpdates(LocationRequest request, LocationCallback callback, Looper looper)`, and `removeLocationUpdates(LocationCallback callback)`). This PR also adds a new `Parcelable` class `LocationAvailability` as well as a new interface, `LocationCallback`.

Closes #64

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/api/LocationAvailability.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/LocationAvailability.java
@@ -1,0 +1,94 @@
+package com.mapzen.android.lost.api;
+
+import android.content.Intent;
+import android.os.Parcel;
+import android.os.Parcelable;
+
+/**
+ * Status on the availability of location data.
+ *
+ * Delivered from LocationCallback registered via
+ * {@link FusedLocationProviderApi#requestLocationUpdates(
+ * LocationRequest,LocationCallback,android.os.Looper) or from a PendingIntent registered via
+ * {@link FusedLocationProviderApi##requestLocationUpdates(LocationRequest,
+ * android.app.PendingIntent). It is also available on demand via
+ * {@link FusedLocationProviderApi#getLocationAvailability()}.
+ */
+public class LocationAvailability implements Parcelable {
+
+  public static final String EXTRA_LOCATION_AVAILABILITY =
+      "com.mapzen.android.lost.EXTRA_LOCATION_AVAILABILITY";
+
+  boolean locationAvailable = false;
+
+  public LocationAvailability(boolean available) {
+    locationAvailable = available;
+  }
+
+  protected LocationAvailability(Parcel in) {
+    locationAvailable = in.readByte() != 0;
+  }
+
+  public static final Creator<LocationAvailability> CREATOR = new Creator<LocationAvailability>() {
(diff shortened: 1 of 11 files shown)


lost/src/main/java/com/mapzen/android/lost/api/LocationAvailability.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override public LocationAvailability[] newArray(int size) {
+      return new LocationAvailability[size];
+    }
+  };
+
+  /**
+   * Extracts the LocationAvailability from an Intent.
+   * @param intent
```
ecgreb: We should start filling in Javadoc descriptions for params and return values where appropriate. Using `@link` and `@code` tags would also be a nice touch. For example:

```java
  /**
   * Extracts the {@link LocationAvailability} from an {@link Intent}.
   * @param intent an {@link Intent} that may or may not include {@link LocationAvailability}.
   * @return a {@link LocationAvailability} object, or {@code null} if the given {@link Intent} does not contain this data.
   */
```

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
```diff
@@ -139,4 +188,25 @@ public void reportProviderEnabled(String provider) {
   public Map<PendingIntent, LocationRequest> getPendingIntents() {
     return intentToRequest;
   }
+
+  private LocationAvailability createLocationAvailability() {
```
ecgreb: I wonder if this method should take the check one step further and verify not only at least one location provider is enabled but also that `LocationManager#getLastKnownLocation(provider)` returns a non-null value for at least one provider?

sarahlensing: I like that idea. Updated.

lost/src/main/java/com/mapzen/android/lost/api/LocationAvailability.java
sarahlensing: Updated javadocs.

PR #86, opened 2016-08-26
Initial geofencing support

ecgreb:
### Overview

This pull request adds basic geofence support. The scope of initial support is to register a single proximity alert using the `GeofencingRequest` API. Under the hood the geofencing API uses [`LocationManager.addProximityAlert(...)`](https://developer.android.com/reference/android/location/LocationManager.html#addProximityAlert(double, double, float, long, android.app.PendingIntent)) to implement the alerts.

### Proposed Changes

Add implementation for `GeofencingApi.addGeofences(GeofencingRequest geofencingRequest, PendingIntent pendingIntent)` to allow client apps to create a proximity alert and receive a `PendingIntent`  callback when the geofence is entered or exited.

Also a new `Activity` and `IntentService` will be added to the sample application to demonstrate the geofencing API.

### TODO

* **At this time proximity alerts are not yet properly triggered in the sample application. More debugging is required.**
* Add support for registering multiple geofences at one time using `addGeofences(List<Geofence> geofences, PendingIntent pendingIntent)`.
* Add support for removing geofences using `removeGeofences(List<String> geofenceRequestIds)`.
* Connect the `GeofencingApi` to Lost's location mocking infrastructure so proximity alerts can be tested using `setMockLocation()` and `setMockTrace()` in the `FusedLocationProviderApi`.

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/api/GeofencingApi.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/GeofencingApi.java
@@ -6,6 +6,8 @@ import java.util.List;
 
 public interface GeofencingApi {
 
+  void addGeofences(GeofencingRequest geofencingRequest, PendingIntent pendingIntent);
+
   void addGeofences(List<Geofence> geofences, PendingIntent pendingIntent);
 
   void removeGeofences(List<String> geofenceRequestIds);
--- a/lost/src/main/java/com/mapzen/android/lost/api/GeofencingApi.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/GeofencingApi.java
@@ -1,16 +1,24 @@
 package com.mapzen.android.lost.api;
 
 import android.app.PendingIntent;
+import android.support.annotation.RequiresPermission;
 
 import java.util.List;
 
+import static android.Manifest.permission.ACCESS_COARSE_LOCATION;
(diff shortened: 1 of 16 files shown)


ecgreb:
Thanks to @friesenkiwi geofences are now working in the sample app. He caught that I was failing to set a duration/expiration for geofences in the sample app and the default value in `Geofence.Builder` was `0`. Now the default is `NEVER_EXPIRE` and the sample app explicitly sets this as well.

I think this now constitutes a good chunk of working code for preliminary geofence support. Once this is merged then https://github.com/mapzen/LOST/pull/88 should be rebased against master. Then we can implement additional geofence functionality in an iterative fashion using https://github.com/mapzen/LOST/pull/88 and other follow-up pull requests.

lost/src/main/java/com/mapzen/android/lost/api/GeofencingApi.java
```diff
@@ -6,6 +6,8 @@
 
 public interface GeofencingApi {
 
+  void addGeofences(GeofencingRequest geofencingRequest, PendingIntent pendingIntent);
```
sarahlensing: We should add this annotation here: `@RequiresPermission(anyOf = {ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION})`

ecgreb: Added

PR #87, opened 2016-08-26
Multiple location clients

### LINKED ISSUE #62: Support multiple location clients
We should explore the feasibility of supporting multiple location clients from multiple threads and/or processes.

This primary use case would be for a single application if the UI thread and a service thread both needed to register for location updates via separate clients.

The secondary and potential future use case would be to allow a shared instance of lost to run in a background service that could be used by more than one application.

sarahlensing:
### Overview
This PR adds support for multiple `LostApiClient`s to be used with the `LocationServices#FusedLocationApi`.

### Proposed Changes
The PR adds `LostApiClient` as a required parameter to the `FusedLocationProviderApi` and implements scoping `LocationListeners`, `PendingIntents`, and `LocationCallbacks` to each `LostApiClient`. It introduces a new class, `ClientManager` for assisting with disconnection logic in the `LostApiClient` so that the static APIs will be set to `null` when there are no more clients in the `ClientManager`. In addition, this PR changes how we handle `LocationEngines` and now creates an engine per client so that setting a mock location for one client does not affect the mock location of another client.

### TODO
- ~~Implement support for multiple clients in remaining `FusedLocationProviderApi` methods~~
- ~~Add tests for `FusedLocationProviderApi` multiple client support~~
- Implement and add tests for `GeofencingApi` ~~and `SettingsApi`~~ multiple client support

### Future PR
- Implement and add tests for `GeofencingApi` multiple client support

Closes #62


lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
```diff
@@ -28,190 +31,320 @@
...
+      PendingIntent callbackIntent) {
+    Map<PendingIntent, LocationRequest> map = intentToRequest.get(apiClient);
+    if (map == null) {
+      map = new HashMap<>();
+    }
+    map.put(callbackIntent, request);
+    intentToRequest.put(apiClient, map);
+
+    LocationEngine locationEngine = clientToLocationEngine.get(apiClient);
     locationEngine.setRequest(request);
```
ecgreb: The fused location sample activity is crashing on startup because`clientToLocationEngine.get(apiClient)` returns `null` for the location engine. It appears when the engine is created it does not get properly registered.

So that brings up two questions:

1. How can we ensure clients and location engines are properly registered?
2. How should we handle the error case when a request is made using client that is not registered?

```java
08-31 11:52:24.935 29271 29271 E AndroidRuntime: FATAL EXCEPTION: main
08-31 11:52:24.935 29271 29271 E AndroidRuntime: Process: com.example.lost, PID: 29271
08-31 11:52:24.935 29271 29271 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mapzen.android.lost.internal.LocationEngine.setRequest(com.mapzen.android.lost.api.LocationRequest)' on a null object reference
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.mapzen.android.lost.internal.FusedLocationProviderServiceImpl.requestLocationUpdates(FusedLocationProviderServiceImpl.java:103)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.mapzen.android.lost.internal.FusedLocationProviderService.requestLocationUpdates(FusedLocationProviderService.java:72)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl.requestLocationUpdates(FusedLocationProviderApiImpl.java:94)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.example.lost.FusedLocationApiActivity.onConnected(FusedLocationApiActivity.java:173)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl$1.onServiceConnected(FusedLocationProviderApiImpl.java:43)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1223)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1240)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at android.os.Handler.handleCallback(Handler.java:739)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:95)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:148)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at android.app.ActivityThread.main(ActivityThread.java:5417)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at java.lang.reflect.Method.invoke(Native Method)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
08-31 11:52:24.935 29271 29271 E AndroidRuntime:        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
08-31 11:52:24.937  1182  2999 W ActivityManager:   Force finishing activity com.example.lost/.FusedLocationApiActivity
08-31 11:52:24.943  1182  2999 W ActivityManager:   Force finishing activity com.example.lost/.MainActivity
```

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
```diff
@@ -28,190 +31,320 @@
 
   private Context context;
 
-  private boolean mockMode;
-  private LocationEngine locationEngine;
-  private Map<LocationListener, LocationRequest> listenerToRequest;
-  private Map<PendingIntent, LocationRequest> intentToRequest;
-  private Map<LocationCallback, LocationRequest> callbackToRequest;
-  private Map<LocationCallback, Looper> callbackToLooper;
+  private Map<LostApiClient, Map<LocationListener, LocationRequest>> listenerToRequest;
```
ecgreb: We may need to rename this class to `CartographyServiceImpl` because of all these maps 😉 

But seriously I wonder if there is a way we can simplify this design. What do you think about expanding the scope of the `ClientManager` to track which `LocationRequests` are associate with which clients?

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderService.java
```diff
@@ -52,56 +58,63 @@ public FusedLocationProviderService getService() {
     Log.d(TAG, "[onDestroy]");
   }
 
-  public Location getLastLocation() {
-    return serviceImpl.getLastLocation();
+  public Location getLastLocation(LostApiClient apiClient) {
```
ecgreb: What do you think about renaming this param from `apiClient` to `client` in all the method signatures? It would be a bit more brief and also consistent with the [play services API](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderApi.html#getLastLocation(com.google.android.gms.common.api.GoogleApiClient)).

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
sarahlensing: :D

I like that idea...I've pushed code to expand the `ClientManager` which moves these maps into that class

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderService.java
sarahlensing: I'm into it. Updated and pushed new name.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderServiceImpl.java
sarahlensing: This has been resolved with commit cc2419301e4aa6cdd77e7cb38045aaf6a05c5a23

lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
```diff
@@ -0,0 +1,252 @@
...
+ * {@link FusedLocationProviderServiceImpl} to manage client's {@link LocationListener}s,
+ * {@link PendingIntent}s, and {@link LocationCallback}s.
+ */
+public class LostClientManager implements ClientManager {
+  private static final String TAG = ClientManager.class.getSimpleName();
+
+  private static LostClientManager instance;
+  private HashSet<LostApiClient> clients;
+
+  private Map<LostApiClient, Map<LocationListener, LocationRequest>> listenerToRequest;
```
ecgreb: Do we actually need to keep track of the `LocationRequest` or would using a `Map<LostApiClient, List<LocationListener>` be sufficient here?

The requests don't appear to be used in this class. And the `LocationEngine` maintains its own list of requests in `LocationRequestUnbundled`.

If the `LocationRequest`s can be safely removed we could simplify the data structures to store `PendingIntent`s and `LocationCallback`s in the same way.

sarahlensing: Nice catch, I safely removed the `LocationRequest` from all the data structures and repushed!

PR #93, opened 2016-09-06
Make LocationServices fields final

### LINKED ISSUE #33: Make FusedLocationApi and GeofencingApi fields final.
Static fields that provide entry point to `LocationServices` should not be able to be modified by the client app.

sarahlensing:
### Overview
This PR prevents the static APIs from being able to be modified by clients.

### Proposed Changes
- Make `LocationServices` fields final
- Add `connect()`, `isConnected()`, and `disconnect()` methods for all APIs
- Update `LostApiClientImpl` to use new `connect()`, `isConnected()`, and `disconnect()` methods for corresponding methods
- Update and add tests

Closes #33

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -32,7 +32,7 @@ public class FusedLocationProviderApiImpl
 
   private static final String TAG = FusedLocationProviderApiImpl.class.getSimpleName();
 
-  private final Context context;
+  private Context context;
   private FusedLocationProviderService service;
   private boolean connecting;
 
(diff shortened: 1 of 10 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -55,11 +55,12 @@
 
   LostApiClient.ConnectionCallbacks connectionCallbacks;
 
-  public FusedLocationProviderApiImpl(Context context) {
-    this.context = context;
+  public FusedLocationProviderApiImpl() {
```
ecgreb: Empty public constructor can be removed

lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
```diff
@@ -15,12 +15,23 @@
  */
 public class GeofencingApiImpl implements GeofencingApi {
 
-  private final LocationManager locationManager;
+  private LocationManager locationManager;
 
-  GeofencingApiImpl(Context context) {
+  public GeofencingApiImpl() {
```
ecgreb: Empty public constructor can be removed

lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
```diff
@@ -29,7 +29,7 @@
 public class LostApiClientImplTest {
   private LostApiClient client;
 
-  @Before public void setUp() throws Exception {
```
ecgreb: Setup and test methods in JUnit should all include `throws Exception` by convention.

This makes the tests less brittle since it ensures that if any method signatures change for the production code under test to throw an exception then the tests will still compile.

PR #97, opened 2016-09-08
Change checkstyle rules to allow for current version

friesenkiwi:
### Overview
Hi, I wanted to use Checkstyle directly from within the Android Studio GUI, so hints can be displayed while coding, I thought that might make it easier to adhere to the style.

The current plugin (https://plugins.jetbrains.com/plugin/1065, 4.31.0, 2016-08-06) comes with Checkstyle 7.1/31 Jul 2016, but the rule `RedundantThrows`, used in `config/checkstyle/checkstyle.xml`
 is not being supported any more by Checkstyle since v. 6.2/28 Dec 2014 (because it is said to be an "Example of really buggy Check", see https://github.com/checkstyle/checkstyle/issues/473#issuecomment-91093541 and http://checkstyle.sourceforge.net/releasenotes.html#Release_6.2).

I tried using an outdated version of the plugin (4.14.3, 2015-03-14, https://plugins.jetbrains.com/update/index?pr=&updateId=18978) with the old Checkstyle, which works, but I guess it would be better to use the current versions and remove the rule, since it is buggy anyways?

Also, it took me some time to get around setting and using code style and Checkstyle rules in Android Studio, so I thought why not write it up for others to come.
I hope, it's not too intrusive for a contributor to change the contributing guidelines...?

### Proposed Changes
- Remove `RedundantThrows` from Checkstyle to allow for latest version
- Add some information to the contributing guidelines

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,6 +2,13 @@
 
 Contributor guidelines for the Location Open Source Tracker project.
 
+First off, thanks for taking the time to contribute!
+
+## Aim
+This project seeks to provide an open source alternative to Google's [Fused Location Provider](https://developer.android.com/training/location/index.html), only depending on the Android SDK.
+Therefore, contributions should implement the [FusedLocationProviderAPI](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderApi) or fix bugs/improve the current implementation, rather than implementing completely new features.
+But a working subset is always fine. That's kind of what this project is all about ;).
+
 ## Prerequisites
 
 1. Android SDK with Tools, Extras, and API 23 installed.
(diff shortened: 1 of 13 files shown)


.github/CONTRIBUTING.md
```diff
@@ -2,60 +2,110 @@
 
 Contributor guidelines for the Location Open Source Tracker project.
 
+First off, thanks for taking the time to contribute!
+
+## Aim
+This project seeks to provide an open source alternative to Google's [Fused Location Provider](https://developer.android.com/training/location/index.html), only depending on the Android SDK.
+Therefore, contributions should implement the [FusedLocationProviderAPI](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderApi) or fix bugs/improve the current implementation, rather than implementing completely new features.
```
ecgreb: I would add here that the [GeoencingApi](https://developers.google.com/android/reference/com/google/android/gms/location/GeofencingApi) and [SettingsApi](https://developers.google.com/android/reference/com/google/android/gms/location/SettingsApi) are also supported in addition to `FusedLocationProviderApi`. 

Basically our goal is to support an open source implementation of anything in the the [com.google.android.gms.location](https://developers.google.com/android/reference/com/google/android/gms/location/package-summary) package.

ecgreb:
This is great. We really appreciate this kind of feedback from the perspective of a contributor and seeing what you think would be helpful for others to get started.

I will have to give the IntelliJ checkstyle plugin a spin now that it works with the project ;)

.github/CONTRIBUTING.md
```diff
@@ -2,60 +2,110 @@
...
+
+## Managing upstream changes
+While you work on your feature branch, there may be changes happening on the upstream master. It is helpful to monitor them, by fetching from time to time.
+Please do not merge them into your branch, but instead [rebase your changes](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) before issuing the pull request.
+Personally we prefer rebase to merge commits since it makes changes easier to review and history a bit more tidy. 
+You can do the rebase from the command line or, in Android Studio, while you are on your branch, select "Rebase onto" from the context menu of the upstream master in the "Log" of the "Version Control" tab.
+You will probably (depending on how big the master changes and your's are) have to "Resolve conflicts" in the files separately (select from the context menu from the individual files shown in red under "Log", always finish by "Merge"), and afterwards select "Skip Commit in Rebasing" from the "VCS->Git" menu.
+More information on this is available at [IntelliJ](https://www.jetbrains.com/help/idea/2016.2/interactive-rebase.html) or [Youtube](https://www.youtube.com/watch?v=lILdMGa-r38).
```
ecgreb: I think we could do without this video link. The other resources provided here seem sufficient. And while the video is somewhat helpful it doesn't really cover rebasing until the very end (after showing how to create merge commits - which is what we want to avoid).

friesenkiwi:
Ok, glad you like it :-)
I included your remarks in the latest commit.
And yes, I think the Checkstyle Plugin makes things much easier. ;-)

ecgreb:
👍

PR #113, opened 2016-09-30
Add upgrade instructions for 1.x to 2.0

sarahlensing:
### Overview
Add instruction page for how to upgrade from 1.x to 2.0


ecgreb:
This looks good. We should also mention that its now on the app developer to include any necessary permissions in their AndroidManifest.xml and perform all runtime permission checks since we took these out of the library.

Maybe we could list each permission that Lost uses and which APIs its needed for as a general guideline.

sarahlensing:
Nice catch. I've updated with those pieces of info

PR #115, opened 2016-10-04
101 transition type

### LINKED ISSUE #101: Implement transition type
### Description

`Geofence.Builder` should allow setting of transition types. Alerts should only be generated for the specified transitions.

sarahlensing:
### Overview
This PR implements checking the `Geofence` transition type to determine if the `GeofencingApi` should fire corresponding `PendingIntents`.

### Proposed Changes
The changes implement transition type in `Geofence` and update `GeofencingApiImpl` to check the value of a `Geofence`'s transition to make sure it includes the type received by the `PendingIntent` generated by `LocatationManager#addProximityAlert` before firing intents. It was necessary to remove all extras added by `GeofencingApiImpl` in order to receive the extra generated by LocationManager ("entering"), they were not merged. Instead of adding `Geofence` and `PendingIntent` to the extras, a map is kept and used when the external intent is constructed in `GeofencingIntentSender`.

Closes #101


lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
```diff
@@ -123,4 +129,16 @@ private void removeGeofences(LostApiClient client, String geofenceRequestId)
...
 
+  public PendingIntent pendingIntentForIntentId(int intentId) {
+    return idToPendingIntent.get(intentId);
+  }
+
+  public Geofence geofenceForIntentId(int intentId) {
+    return idToGeofence.get(intentId);
+  }
+
+  public void setGeofenceForIntentId(int intentId, Geofence geofence) {
```
ecgreb: It would be great if we could find a way to test this without adding a public method that is only called from tests. Is it possible to just have the tests call `addGeofence(...)`?

lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
```diff
@@ -70,12 +72,16 @@ public void disconnect() {
   private PendingResult<Status> addGeofence(LostApiClient client, Geofence geofence,
       PendingIntent pendingIntent) throws SecurityException {
 
-    internalIntent = intentFactory.createIntent();
-    internalIntent.putExtra(GeofencingIntentService.EXTRA_PENDING_INTENT, pendingIntent);
+    int pendingIntentId = (int) System.currentTimeMillis();
```
sarahlensing: We could but it might not be worth it because this id needs to be an integer. It's used when we call `PendingIntent getService(Context context, int requestCode, Intent intent, int flags)`. The `geofence.getRequestId()` is a string so we'd have to check for an exception when converting the string to an integer.

lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
```diff
@@ -70,12 +72,16 @@ public void disconnect() {
...
+    internalIntent = intentFactory.createIntent(context);
+    internalIntent.addCategory(String.valueOf(pendingIntentId));
```
sarahlensing: When we added extras to the `internalIntent`, the "entering" extra that the `LocationManager` was supposed to add wasn't present when the intent fired :/

lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
sarahlensing: Rm'd this public method and extracted the id generation into a separate class that implements the new `IdGenerator` interface

PR #119, opened 2016-10-05
116 dwell transition

### LINKED ISSUE #116: Produce DWELL transition events

sarahlensing:
### Overview
This PR adds support for setting a loitering delay on `Geofence` as well as dispatching dwell transition events.

### Proposed Changes
When a geofence is entered, an alarm is scheduled to fire after a delay determined by the value of the geofence's loitering delay. When the alarm fires, the dwell transition is sent in the `Intent` data. If a geofence is exited before this interval, then the alarm is cancelled and no dwell transition is fired.

Closes #116

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/DwellServiceIntentFactory.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/DwellServiceIntentFactory.java
@@ -16,6 +16,6 @@ public class DwellServiceIntentFactory implements IntentFactory {
 
   @Override
   public PendingIntent createPendingIntent(Context context, int pendingIntentId, Intent intent) {
-    return PendingIntent.getService(context, 0, intent, 0);
+    return PendingIntent.getService(context, pendingIntentId, intent, 0);
   }
 }
--- a/lost/src/main/java/com/mapzen/android/lost/internal/DwellServiceIntentFactory.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/DwellServiceIntentFactory.java
@@ -1,0 +1,21 @@
+package com.mapzen.android.lost.internal;
+
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+
+/**
+ * Class to handle creating {@link Intent} and {@link PendingIntent} objects to be sent to
+ * {@link DwellIntentService}
+ */
+public class DwellServiceIntentFactory implements IntentFactory {
+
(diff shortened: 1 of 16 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/DwellServiceIntentFactory.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public class DwellServiceIntentFactory implements IntentFactory {
+
+  @Override public Intent createIntent(Context context) {
+    return new Intent(context, DwellIntentService.class);
+  }
+
+  @Override
+  public PendingIntent createPendingIntent(Context context, int pendingIntentId, Intent intent) {
+    return PendingIntent.getService(context, 0, intent, 0);
```
ecgreb: Should we be sending the `pendingIntentId` passed into this method instead of `0`?

lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
```diff
@@ -74,16 +82,18 @@ public void disconnect() {
   private PendingResult<Status> addGeofence(LostApiClient client, Geofence geofence,
```
ecgreb: I noticed the `client` param is unused when we add a geofence. We should probably be associating geofences with the client on which they were requested the same way we are with location update requests?

This way if one client is shutdown we can remove all its associated geofences but not impact those registered by another client.

Anyway just a thought that came to mind when reading this code. We don't have to deal with it in this change set but we should probably create an issue to handle at a later time.

ecgreb:
We should probably also add transition type and loitering delay to the sample app to facilitate manual testing for dwell events.

lost/src/main/java/com/mapzen/android/lost/internal/DwellServiceIntentFactory.java
sarahlensing: Yes! Fixed.

lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
sarahlensing: Nice catch, that will be a nice improvement.

Tracking here: https://github.com/mapzen/lost/issues/120

sarahlensing:
Updated the sample app to include these

PR #127, opened 2016-10-11
Fix mock trace tests and sample app

### LINKED ISSUE #84: Fix tests using Thread.sleep(100)
### Description

A number of tests are currently ignored due to intermittent failures. These tests all depend on concurrency used by `MockEngine` to deliver mock location updates from a GPX trace file. We need to find a better way to test the behavior of `FusedLocationProviderApi.setMockTrace(File)` without the use of `Thread.sleep(100)`.

### Steps to Reproduce

Remove all `@Ignore("Intermittently failing. Find a better way to test without Thread.sleep(100)")` annotations and run tests.

### Lost & Android Version

LOST 1.1.1

ecgreb:
### Overview

* Fixes mock trace unit tests that would fail due to concurrency issues.
* Fixes mock routing in sample app.

### Proposed Changes

Refactors mock GPX trace logic so it can be tested synchronously. Cleans up classes and adds interfaces to facilitate testing.

Updates sample app to load a mock GPX trace from its external files directory. Previously the trace file was loaded from the root of the SD card which now fails due to insufficient permissions.

This is cleaner than adding the required permissions for accessing all of external storage and is the now the recommended way for developers to use mock traces in a runtime permissions world.

Also adds `fastestInterval` to sample app location request since this is required for mock trace replay to work correctly (background thread sleeps for `fastestInterval` between each location update).

Adds utility script `install-gpx-trace.sh` to make it easier to copy a GPX trace file into an application's external files directory on an emulator or device.

Fixes #84

### DIFF
--- a/docs/mock-locations-routes.md
+++ b/docs/mock-locations-routes.md
@@ -1,8 +1,8 @@
-#Mock Locations
+# Mock Locations
 
 With Lost you can mock not just individual locations but also entire routes. By loading a [GPX trace file](http://www.topografix.com/gpx.asp) onto the device you can configure Lost to replay locations from the trace file including latitude, longitude, speed, and bearing.
 
-**Mocking a single location**
+## Mocking a single location
 To mock a single location with Lost you must first enable mock mode. Then you simply create a mock location object and pass it to the API.
 
 ```java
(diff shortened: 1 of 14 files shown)


sarahlensing:
These changes are great. We might want to update the mock-locations-routes.md instructions to add your new script. Maybe something like:

```markdown
# Install GPX File
To install your gpx file onto your device, run the install script included in this repo:
install-gpx-trace.sh <trace file> <package name>

Example:
install-gpx-trace.sh lost.gpx com.example.myapp
```

ecgreb:
@sarahlensing good call. Updated docs to include gpx install script.

https://github.com/mapzen/lost/pull/127/commits/577396a5cbb659cd30a088eee3c8c7ee2fb1586e

PR #128, opened 2016-10-12
Dont require bluetooth permission

### LINKED ISSUE #121: Dont require bluetooth permission for SettingsApi
### Description
Update `LocationSettingsResultRequest` to check that user has declared bluetooth permission before when calculating `bleUsable` and before calling `BluetoothAdapter#isEnabled()`

https://github.com/mapzen/lost/blob/a5a2d9fa687f43a6d4aee43f2118114b4343a046/lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java#L112

sarahlensing:
### Overview
Don't require bluetooth permission when using the `SettingsApi`

### Proposed Changes
Check that the bluetooth permission is granted before calling `bluetoothAdapter.isEnabled()`

Closes #121

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
@@ -8,8 +8,10 @@ import com.mapzen.android.lost.api.PendingResult;
 import com.mapzen.android.lost.api.ResultCallback;
 import com.mapzen.android.lost.api.Status;
 
+import android.Manifest;
 import android.app.PendingIntent;
 import android.bluetooth.BluetoothAdapter;
+import android.content.Context;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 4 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/LocationSettingsResultRequest.java
```diff
@@ -40,8 +43,10 @@
 
   Future<LocationSettingsResult> future;
 
-  public LocationSettingsResultRequest(BluetoothAdapter btAdapter, PackageManager pm,
-      LocationManager lm, PendingIntentGenerator generator, LocationSettingsRequest request) {
+  public LocationSettingsResultRequest(Context context, BluetoothAdapter btAdapter,
```
ecgreb: To avoid adding more constructor params we should use the `Context` to get a reference the `PackageManager` rather than passing both.

PR #130, opened 2016-10-18
Fix connection state

sarahlensing:
### Overview
This PR updates the internal connection state for the `FusedLocationProviderApi` before invoking any callbacks. Previously, callbacks were invoked before the state was updated which caused bugs when client logic in callbacks relied on checking `LostApiClient#isConnected`.

### Proposed Changes
This work extracts non-service related logic from `FusedLocationProviderApiImpl` into a new class, `FusedLocationServiceConnectionManager` and adds test coverage for the new class.

Resolves problem described here: https://github.com/mapzen/android/issues/208

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -34,7 +34,7 @@ public class FusedLocationProviderApiImpl
 
   private Context context;
   private FusedLocationProviderService service;
-  private enum ConnectState { IDLE, CONNECTING, CONNECTED };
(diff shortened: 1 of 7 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -30,95 +28,81 @@
...
-      }
-      connectState = ConnectState.IDLE;
-      if (stopService) {
-        context.unbindService(serviceConnection);
-        Intent intent = new Intent(context, FusedLocationProviderService.class);
-        context.stopService(intent);
-        service = null;
-      }
-    }
+    serviceConnectionManager.disconnect(client, stopService);
```
sarahlensing: I don't have an answer to that- removing it as you're right, its only ever called with `true`

PR #135, opened 2016-10-24
Better is connected logic for multiple clients

### LINKED ISSUE #134: Update LostApiClient#isConnected logic
The client should consider whether or not it is bound

sarahlensing:
### Overview
This PR updates how `LostApiClient#isConnected` is determined to consider whether the client is being managed by the `ClientManager`.

### Proposed Changes
This fixes a bug where when multiple clients were connected, the method would report that unbound clients were connected.

Closes #134

### DIFF
--- a/lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
+++ b/lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
@@ -166,7 +166,8 @@ public class LostApiClientImplTest {
     assertThat(client.isConnected()).isFalse();
   }
 
-  @Test public void isConnected_multipleClients_shouldReturnFalseAfterDisconnected() throws Exception {
+  @Test public void isConnected_multipleClients_shouldReturnFalseAfterDisconnected() throws
+      Exception {
     LostApiClient anotherClient = new LostApiClient.Builder(application).build();
     anotherClient.connect();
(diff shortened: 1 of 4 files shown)


lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
```diff
@@ -166,4 +165,15 @@ public void isConnected_shouldReturnFalseBeforeConnected() throws Exception {
...
   }
+
+  @Test public void isConnected_multipleClients_shouldReturnFalseAfterDisconnected() throws
+      Exception {
+    LostApiClient anotherClient = new LostApiClient.Builder(application).build();
+    anotherClient.connect();
+    client.connect();
+    client.disconnect();
+    assertThat(client.isConnected()).isFalse();
+    anotherClient.disconnect();
```
ecgreb: What is the purpose of calling `anotherClient.disconnect()` after the assert? Is this necessary? Should there be another assert after it?

sarahlensing: Its needed so that the shared `LostClientManager` is cleaned up and the tests in `LostClientManagerTest` run as expected. Though maybe we want to refactor `LostApiClientImpl` to inject the `ClientManager` in the constructor?

PR #137, opened 2016-11-03
Location provider enabled/disabled

### LINKED ISSUE #123: onProviderDisabled/Enabled is not working properly
- When gps provider status changed, the onProviderDisabled/Enabled is working properly.
- But when network provider  status changed, the onProviderDisabled/Enabled is not working properly。
- (I close wifi , mobile data, change to airplane mode.  the onProviderDisabled/Enabled is not work.)

sarahlensing:
### Overview
This PR changes the underlying mechanism that Lost uses to notify of providers enabled/disabled from a `LocationListener` to a `ContentObserver` on the system location provider setting.

### Proposed Changes
`FusionEngine` no longer does anything with its provider changed `LocationListener` methods and instead uses `LocationProviderChangedHandler` to receive notifications of providers enabled & disabled. The `LocationProviderChangedHandler` handles setting up a `ContentObserver` to listen to changes to `Settings.System.LOCATION_PROVIDERS_ALLOWED`.

Closes #123

### DIFF
--- a/lost-sample/src/main/java/com/example/lost/LocationAvailabilityActivity.java
+++ b/lost-sample/src/main/java/com/example/lost/LocationAvailabilityActivity.java
@@ -1,0 +1,60 @@
+package com.example.lost;
+
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v7.app.AppCompatActivity;
+import android.view.View;
+import android.widget.Button;
+import android.widget.Toast;
+
+import com.mapzen.android.lost.api.LocationAvailability;
+import com.mapzen.android.lost.api.LocationServices;
+import com.mapzen.android.lost.api.LostApiClient;
+
+public class LocationAvailabilityActivity extends AppCompatActivity {
+
+    LostApiClient client;
+
+    @Override protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_location_availability);
+
+        setupLocationAvailabilityBtn();
+    }
+
+    private void setupLocationAvailabilityBtn() {
+        Button connect = (Button) findViewById(R.id.check_availability);
+        connect.setOnClickListener(new View.OnClickListener() {
+            @Override public void onClick(View v) {
+                if (client == null || !client.isConnected()) {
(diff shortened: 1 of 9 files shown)


lost/src/main/java/com/mapzen/android/lost/api/LocationListener.java
```diff
@@ -1,6 +1,7 @@
 package com.mapzen.android.lost.api;
 
 import android.location.Location;
+import android.os.Looper;
```
ecgreb:
```
:lost:checkstyle[ant:checkstyle] /home/ubuntu/lost/lost/src/main/java/com/mapzen/android/lost/api/LocationListener.java:4:8: Unused import - android.os.Looper.
```

lost-sample/src/main/java/com/example/lost/LocationAvailabilityActivity.java
```diff
@@ -0,0 +1,60 @@
...
+            public void onConnectionSuspended() {
+
+            }
+        }).build();
+        client.connect();
+    }
+
+    private void checkLocationAvailability() {
+        LocationAvailability availability =
+                LocationServices.FusedLocationApi.getLocationAvailability(client);
```
ecgreb: While exercising a call to `FusedLocationApi.getLocationAvailability(client)` is good we should also verify that `LocationCallback` is properly invoked when location updates have been requested and provider status changes occur.

https://github.com/mapzen/lost/blob/master/lost/src/main/java/com/mapzen/android/lost/api/FusedLocationProviderApi.java#L100-L101

This is the use case that would more suitably replace `LocationListener.onProviderEnabled(String provider)` and `LocationListener.onProviderDisabled(String provider)` since the client app is relying on a push style notification rather than actively polling the provider status.

lost/src/main/java/com/mapzen/android/lost/api/LocationListener.java
```diff
@@ -17,14 +18,28 @@
   void onLocationChanged(Location location);
 
   /**
-   * Called when a location provider is disabled.
+   * Called when a location provider is disabled. You will only receive updates for the priority
```
ecgreb: :+1:

PR #144, opened 2016-11-28
Remove connection callbacks when client disconnected

### LINKED ISSUE #138: onConnected ConnectionCallback is called multiple times
### Description
onConnected ConnectionCallback is called multiple times if you recreate the activity (rotate the screen).

The expectation is that it is just called once because in the onConnected() callback, usually the app requests location updates. In the same breath, of course, since Android 5.0, the permission must be requested first. But if onConnected() is called several times, then the app also requests the permission multiple times.

### Steps to Reproduce

Have an activity with this code. Then, see that the activity is recreated (rotate the screen). Look at the log. The log is logged twice.
If you rotate the screen once more, it will be logged three times. Etc.
```
private LostApiClient lostApiClient;
private int times;

@Override public void onCreate(Bundle savedInstanceState)
{
	super.onCreate(saveInstanceState);
	times = 0;
	lostApiClient = new LostApiClient.Builder(this).addConnectionCallbacks(
		new LostApiClient.ConnectionCallbacks()
		{
			@Override
			public void onConnected()
			{
				Log.i("LOST","onConnected() called " + (++times) + " now");
			}

			@Override
			public void onConnectionSuspended() {}
		}).build();
}

@Override public void onStart()
{
	super.onStart();
	lostApiClient.connect();
}

@Override public void onStop()
{
	super.onStop();
	lostApiClient.disconnect();
}
```

### Lost & Android Version
2.1.1, Android 6.0

sarahlensing:
### Overview
This PR fixes a bug which caused `ConnectionCallbacks#onConnected` to be called multiple times when the lifecycle of the `LostApiClient` was tied to the `Activity` lifecycle.

### Proposed Changes
This adds a call to remove the `LostApiClient` `ConnectionCallback` when the client is disconnected, making the management of callbacks symmetrical. Previously, they were only added when the client was connected.

Closes #138

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -165,11 +165,11 @@ public class FusedLocationProviderApiImpl
     return service;
   }
 
-  public void removeConnectionCallbacks(LostApiClient.ConnectionCallbacks callbacks) {
+  void removeConnectionCallbacks(LostApiClient.ConnectionCallbacks callbacks) {
     serviceConnectionManager.removeCallbacks(callbacks);
   }
 
-  public FusedLocationServiceConnectionManager getServiceConnectionManager() {
+  FusedLocationServiceConnectionManager getServiceConnectionManager() {
     return serviceConnectionManager;
   }
 }
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -164,4 +164,12 @@ public class FusedLocationProviderApiImpl
   public FusedLocationProviderService getService() {
     return service;
   }
+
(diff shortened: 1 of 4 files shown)


lost/src/test/java/com/mapzen/android/lost/internal/LostApiClientImplTest.java
```diff
@@ -13,10 +17,6 @@
 import org.robolectric.RobolectricGradleTestRunner;
 import org.robolectric.annotation.Config;
 
-import android.content.ComponentName;
```
ecgreb: Using the `MapzenAndroid` code style android imports should stay below mapzen and third party.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -164,4 +164,12 @@ public boolean isConnected() {
   public FusedLocationProviderService getService() {
     return service;
   }
+
+  public void removeConnectionCallbacks(LostApiClient.ConnectionCallbacks callbacks) {
```
ecgreb: We should make `removeConnectionCallbacks(...)` and `getServiceConnectionManager()` package private since they are only used internally.

PR #145, opened 2016-12-01
Fixes SettingsApi for BLE location

### LINKED ISSUE #139: Call to SettingsApi.checkLocationSettings requires Bluetooth permission
### Description
`SettingsApi.checkLocationSettings` is annotated with `@RequiresPermission(anyOf = {BLUETOOTH, BLUETOOTH_ADMIN})` which makes Android Studio suggest to add the Bluetooth permission to the manifest, regardless of if the LocationSettingsRequest contains a `.setNeedBle(true)` or not. 

Since setting the `.setNeedBle(true)` is something the user of this library would do deliberately and the default is false, I think it makes sense to remove this annotation and instead document `@throws SecurityException if the LocationSettingsRequest is set to need bluetooth but the app did not declare this permission in the manifest` or alike.

### Lost & Android Version
2.1.1 on Android 6

ecgreb:
### Overview

Previously the `SettingsApi` was relying on `BluetoothAdapter.isEnabled()` to verify if bluetooth location is usable. This was incorrect as BLE location availability is independent from bluetooth communication.

### Proposed Changes

Removes dependency on `BluetoothAdapter.isEnabled()` and instead reports bluetooth location is usable when network location is usable under "High accuracy" or "Battery saving" location settings. This change also removes the requirement for `BLUETOOTH` or `BLUETOOTH_ADMIN` to use the `SettingsApi`.

Also notice how "Bluetooth" has been added to the descriptions of "High accuracy" and "Battery saving" location modes in Android 7.x.

Fixes #139


lost/src/main/java/com/mapzen/android/lost/internal/PendingIntentGenerator.java
```diff
@@ -5,21 +5,16 @@
...
+class PendingIntentGenerator {
 
   private final Context context;
 
-  public PendingIntentGenerator(Context context) {
+  PendingIntentGenerator(Context context) {
     this.context = context;
   }
 
   public PendingIntent generatePendingIntent(boolean hasBleResolution) {
```
sarahlensing: Looks like this parameter can be removed

ecgreb: @sarahlensing good catch. removed.

PR #146, opened 2016-12-02
Respsect LocationRequest Parameters

### LINKED ISSUE #142: onLocationChanged is called from a LocationRequest of another fragment
### Description
Not sure if this is expected behavior (=mirroring google play api) or a bug. In any case, it would be good if this was documented:
Requesting different location updates from different places will make the `onLocationChanged` method be called in both places whenever the condition of either location request of any of the two places is fulfilled.

### Steps to Reproduce
I have an activity which requests location updates only every few minutes and only if the user moved quite a bit.
```
LocationRequest request = LocationRequest.create()
		.setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)
		.setSmallestDisplacement(500)
		.setInterval(3*60*1000); // 3 minutes

LocationServices.FusedLocationApi.requestLocationUpdates(lostApiClient, request, this);
```

Then, I also got a fragment, which creates an own lostApiClient for itself (which I guess points to the same object internally) and requests very frequent location updates, let's say every second.

As soon as that fragment starts requesting location updates, the former mentioned activity starts to get calls every second to `onLocationChanged`.
The activity is doing an expensive operation on `onLocationChanged`, I expected that the parameters of my location request would be respected.

### Lost & Android Version
2.1.1 on Android 5.0

sarahlensing:
### Overview
This PR adds support for respecting `LocationRequest#getFastestInterval()` and `LocationRequest#getSmallestDisplacement()` when deciding whether to notify listeners/pending intents/callbacks.

### Proposed Changes
Now listeners/pending intents/callbacks are invoked if the elapsed time between location updates is greater than/equal to the fastest interval and the distance between the last location reported and the current location is greater than/equal to the smallest displacement in meters.

Closes #142


lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
```diff
@@ -269,4 +289,75 @@ public void shutdown() {
...
+              notifier.notify(client, obj);
+            }
+          }
+        }
+      }
+    }
+    return new ReportedChanges(updatedRequestToReportedTime, updatedRequestToReportedLocation);
+  }
+
+  abstract class Notifier<T> {
```
ecgreb: Why not make `Notifier` an interface since there is no implementation here?

lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
```diff
@@ -269,4 +289,75 @@ public void shutdown() {
...
+    abstract void notify(LostApiClient client, T obj);
```
ecgreb: 👍 

I really like the use of generics to allows us to treat `LocationListener`, `PendingIntent`, and `LocationCallback` objects in a similar fashion when sending location updates.

lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
```diff
@@ -38,14 +38,26 @@
   private Map<LostApiClient, Set<LocationCallback>> clientToLocationCallbacks;
   private Map<LostApiClient, Map<LocationCallback, Looper>> clientCallbackToLoopers;
 
+  private Map<LocationListener, LocationRequest> listenerToLocationRequests;
+  private Map<PendingIntent, LocationRequest> intentToLocationRequests;
+  private Map<LocationCallback, LocationRequest> callbackToLocationRequests;
+  private Map<LocationRequest, Long> requestToLastReportedTime;
+  private Map<LocationRequest, Location> requestToLastReportedLocation;
```
sarahlensing: Replaced

lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
sarahlensing: Good call, changed to an interface

lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
sarahlensing: 😄

sarahlensing:
Yep, I did test those things in a dummy app. I'll update our samples to include tests for these now.

lost-sample/src/main/java/com/example/lost/MultipleLocationListenerSingleClientActivity.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Demonstrates one {@link LostApiClient}s receiving location updates at difference intervals
+ */
+public class MultipleLocationListenerSingleClientActivity extends ListActivity {
+
+  private static final int LOCATION_PERMISSION_REQUEST = 1;
+
+  LostApiClient lostApiClient;
+  LostApiClient otherLostApiClient;
```
ecgreb: This is the single client example but there are still two clients here?

Given the multiple client activity/fragment example I think it would be more useful to test a single instance of the location client that makes multiple location requests here. What do you think?

sarahlensing: Drp...fixed.

PR #149, opened 2016-12-07
Initialize thread as looper if needed

### LINKED ISSUE #131: Can't request location updates via pending intent from non-ui thread
### Description

Request location updates via pending intent from a service, from non-ui thread fails.
### Steps to Reproduce

Simply request location updates from a non-main thread.
### Lost & Android Version

2.1.1, Android 7.0 (24)
### Log

java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
                                                                                         at android.os.Handler.<init>(Handler.java:200)
                                                                                         at android.os.Handler.<init>(Handler.java:114)
                                                                                         at android.location.LocationManager$ListenerTransport$1.<init>(LocationManager.java:239)
                                                                                         at android.location.LocationManager$ListenerTransport.<init>(LocationManager.java:239)
                                                                                         at android.location.LocationManager.wrapListener(LocationManager.java:871)
                                                                                         at android.location.LocationManager.requestLocationUpdates(LocationManager.java:884)
                                                                                         at android.location.LocationManager.requestLocationUpdates(LocationManager.java:471)
                                                                                         at com.mapzen.android.lost.internal.FusionEngine.enableNetwork(FusionEngine.java:133)
                                                                                         at com.mapzen.android.lost.internal.FusionEngine.enable(FusionEngine.java:107)
                                                                                         at com.mapzen.android.lost.internal.LocationEngine.setRequest(LocationEngine.java:45)
                                                                                         at com.mapzen.android.lost.internal.FusedLocationProviderServiceImpl.requestLocationUpdates(FusedLocationProviderServiceImpl.java:65)
                                                                                         at com.mapzen.android.lost.internal.FusedLocationProviderService.requestLocationUpdates(FusedLocationProviderService.java:74)
                                                                                         at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl.requestLocationUpdates(FusedLocationProviderApiImpl.java:127)
                                                                                         at *.BackgroundLocationService.requestUpdates(BackgroundLocationService.java:168)

sarahlensing:
### Overview
Allow requesting location updates from a background thread without requiring the thread to initialize itself as a looper and setup the message queue to run in it.

### Proposed Changes
This PR prepares the current thread to request location updates (like a certain other location library does).

Closes #131

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusionEngine.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusionEngine.java
@@ -7,6 +7,7 @@ import android.location.Location;
 import android.location.LocationListener;
 import android.location.LocationManager;
 import android.os.Bundle;
+import android.os.Looper;
 import android.util.Log;
 
 import java.util.List;
@@ -74,6 +75,11 @@ public class FusionEngine extends LocationEngine implements LocationListener {
   }
 
   @Override protected void enable() {
+    boolean prepLooper = Looper.myLooper() == null;
+    if (prepLooper) {
+      Looper.prepare();
+    }
+
     long networkInterval = Long.MAX_VALUE;
     long gpsInterval = Long.MAX_VALUE;
     long passiveInterval = Long.MAX_VALUE;
@@ -111,6 +117,10 @@ public class FusionEngine extends LocationEngine implements LocationListener {
     }
     if (passiveInterval < Long.MAX_VALUE) {
       enablePassive(passiveInterval);
+    }
+
+    if (prepLooper) {
+      Looper.loop();
     }
   }
 
--- a/lost-sample/src/main/AndroidManifest.xml
+++ b/lost-sample/src/main/AndroidManifest.xml
@@ -34,6 +34,7 @@
     <activity android:name=".SettingsApiActivity"/>
     <activity android:name=".LocationListenerActivity"/>
     <activity android:name=".PendingIntentActivity"/>
+    <activity android:name=".BackgroundPendingIntentActivity"/>
     <activity android:name=".LocationAvailabilityActivity"/>
 
     <service
--- a/lost-sample/src/main/java/com/example/lost/BackgroundPendingIntentActivity.java
+++ b/lost-sample/src/main/java/com/example/lost/BackgroundPendingIntentActivity.java
@@ -1,0 +1,71 @@
+package com.example.lost;
+
+import com.mapzen.android.lost.api.LocationRequest;
+import com.mapzen.android.lost.api.LocationServices;
+
+import android.app.PendingIntent;
+import android.content.Intent;
+import android.os.AsyncTask;
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.view.View;
+import android.widget.Button;
+
+/**
+ * Requests location updates via {@link PendingIntent} from a background thread
+ */
+public class BackgroundPendingIntentActivity extends LostApiClientActivity {
+
+  @Override protected void onCreate(@Nullable Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.activity_pending_intent);
+
+    setupConnectBtn();
+    setupRequestBtn();
+    setupDisconnectBtn();
+  }
+
+  private void setupConnectBtn() {
+    Button connect = (Button) findViewById(R.id.connect);
+    connect.setOnClickListener(new View.OnClickListener() {
+      @Override public void onClick(View v) {
+        connect();
+      }
+    });
+  }
+
+  private void setupRequestBtn() {
+    Button requestUpdates = (Button) findViewById(R.id.request_updates);
+    requestUpdates.setOnClickListener(new View.OnClickListener() {
+      @Override public void onClick(View v) {
+        requestLocationUpdates();
+      }
+    });
+  }
+
+  private void setupDisconnectBtn() {
+    Button disconnect = (Button) findViewById(R.id.disconnect);
+    disconnect.setOnClickListener(new View.OnClickListener() {
+      @Override public void onClick(View v) {
+        disconnect();
+      }
+    });
+  }
+
+  private void requestLocationUpdates() {
+    AsyncTask task = new AsyncTask() {
+      @Override protected Object doInBackground(Object[] objects) {
+        LocationRequest request = LocationRequest.create();
+        request.setPriority(LocationRequest.PRIORITY_BALANCED_POWER_ACCURACY);
+        request.setInterval(100);
+
+        Intent intent = new Intent(PendingIntentService.ACTION);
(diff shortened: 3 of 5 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/FusionEngine.java
```diff
@@ -112,6 +118,10 @@ public FusionEngine(Context context, Callback callback) {
     if (passiveInterval < Long.MAX_VALUE) {
       enablePassive(passiveInterval);
     }
+
+    if (prepLooper) {
+      Looper.loop();
```
ecgreb: I was going to suggest moving `Looper.loop()` into the `if` block at the top of this method along with `Looper.prepare()` however when I tested that change it the location updates were never delivered to the service. Any idea why this needs to be after the location requests are created?

ecgreb: Ok but I guess its still not clear to me why registering for location updates would fail if `Looper.loop()` has already been called for a background thread.

I'm assuming `Looper.loop()` gets called for the main thread as part of app startup and definitely before we would register for location updates and in that scenario it works.

sarahlensing: In this case the code that registers for location updates is never reached because `Looper.loop()` doesn't return.

sarahlensing: On second thought, this approach is flawed because any code in the background thread called after `FusedLocationProviderApi#requestLocationUpdates` will not be executed. Maybe we can simplify the solution to this problem and just run the call to register for location updates with the `LocationManager` on the main thread if we detect there is no looper in the calling thread. Can you think of any downside to this? Also, after further testing, it appears that other libraries don't create a looper for a given thread if there isn't one.

ecgreb: Moving the request to the main thread would not solve #131 entirely since as @amayatsky pointed out location updates would still stop if the app is killed by the system.

The root of the problem with this is that the framework's [FusionEngine](https://github.com/android/platform_frameworks_base/blob/master/packages/FusedLocation/src/com/android/location/fused/FusionEngine.java) runs inside system service and never shuts down and so updates via `PendingIntent` can be registered to wake up a service even if the app has been killed by the system.

The Lost [FusionEngine](https://github.com/mapzen/lost/blob/master/lost/src/main/java/com/mapzen/android/lost/internal/FusionEngine.java) on the other hand runs in a local service and so is tied to the application lifecycle.

Also we should do an explicit test of play services to see what happens if you call `requestLocationUpdates(...)` from a thread that has not called `Looper.prepare()`. Also I am wondering what play services does if updates are requested via `PendingIntent` and the app is killed by the user (rather than the system).

Really there are two issues here:
1. Should it be possible to request location updates from a thread that has not called `Looper.prepare()`?
2. Is there a way to ensure location updates via `PendingIntent` continue to fire even after the application has been killed by the system.

amayatsky: >the app is killed by the user
If we are talking force stop here, then on the Android 3.1+ the app won't receive any intents after that.
>FusionEngine on the other hand runs in a local service
As far as I can see there is a LocationManager under the hood. Maybe you should simply wrap up existing calls like this [void requestLocationUpdates (long minTime, 
                float minDistance, 
                Criteria criteria, 
                PendingIntent intent)](https://goo.gl/W40jYS)?

sarahlensing: @ecgreb Thats true, while thinking about how to solve the exception I lost sight of that use case.

I think that:
1. Developers should be able to request location updates from a thread that hasn't called `Looper.prepare()` because they are able to do so with play services.
2. We could achieve this but it would require that we update the way we handle starting/stopping our service. Right now its a bound service but if we changed it to be an explicitly started/stopped service, it would continue to run in the background even after the calling application is killed. I remember we had this (in addition to the binding) but removed it...maybe we revisit adding it back?

ecgreb: @sarahlensing we do still explicitly start/stop (and bind) the service at this time. However I'm not sure how using both start/stop and bind effects the lifecycle exactly.

https://github.com/mapzen/lost/blob/master/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java#L40-L41

Seems like some more investigation around how the various methods of starting/stopping/binding the service behaves and also how play services behaves under the following conditions would be good:

1. App is backgrounded by user with system home button
2. App is exited by user with system back button
3. App is quit by user using recent apps menu swipe
4. App is force quit by user using "App info > Force Stop"
5. App is killed by system (is there a way to simulate/induce this behavior besides the above?)

ecgreb: @amayatsky Lost does indeed use the `LocationManager` under the hood. The `FusionEngine` is wrapped in a started service and registered for updates as a `LocationListener`.

If we instead wrap the `FusionEngine` in an `IntentService` and register it with the `LocationManager` via `PendingIntent` this could potentially help solve our issue of background use cases when the app is stopped.

sarahlensing: Using this test project: https://github.com/sarahlensing/gps-test

...I observed these results...

Play services:
1. Intents are delivered 
2. Intents are delivered 
3. Intents are delivered 
4. App service is killed without callbacks to `onDestroy` fired, intents stop being delivered (I think this accurately replicates what happens when the system kills it)

I want to do some more debugging around case 4 to understand what is happening. Initially, I thought that the play service was restarted. Using `adb shell dumpsys activity services FusedLocationService` I can see the play service is still running. And then looking at details dumped I saw that `createTime=-19h33m35s710ms` changed from what it was before I force quit the app. However, I dont think this is what it appears to be because it also changed after I quit the app from the recent menu but intents still continued to be delivered. There is something else going on in this case but I'm not sure what.

LOST:
1. Intents are delivered
2. Intents are delivered
3. App service is killed, Lost service is restarted, intents stop being delivered
4. App and Lost services are killed, intents stop being delivered

I'm going to try starting the Lost service in a different process to see if that changes any of the results.

ecgreb: Thanks for the detailed analysis @sarahlensing. So it looks like the main inconsistency is for *3. App is quit by user using recent apps menu swipe*

And then there is the issue of when the app is killed by the system which may not be the same as a user force quit honestly I am not sure.

Another nice way to get visibility into what services are running is under system settings *Developer options > Running services*

ecgreb:
Based on the discussion it seems like we need to rethink our approach on this one. Closing for now but let's reopen if we find a better way to deal with pending intent update requests from a background thread.

PR #150, opened 2016-12-12
Display alert to optionally fire location settings activity

### LINKED ISSUE #140: LocationServices.SettingsApi seems to be broken
### Description
The location settings api seems to be broken or at least does not work [as documented](https://github.com/mapzen/lost/blob/master/docs/location-settings.md) nor as documented for the Google Play Api (since it should mirror the behavior).
Following points:

1. If the location settings are already on and `checkLocationSettings` is called, the status code returned is `Status.SETTINGS_CHANGE_UNAVAILABLE` instead of `Status.SUCCESS`
2. If the location settings are off and the user is led to the settings activity to turn on the location via `status.startResolutionForResult`, it seems to be unachievable to return to this activity via `onActivityResult` with result code `Activity.RESULT_OK` (see reproduction below)
3. If the location settings are off, the Google Maps app shows a dialog _"To continue, let your device turn on location using Google's location service."_ in which the user can directly turn it on without being led to the settings activity. I am just assuming here that this is done with the Google Play Apis, I never used them before so the assumption could be wrong and this point invalid then. In LOST, calling `status.startResolutionForResult` leads the user directly to the settings, without asking first. Since LOST should mirror Google Play Apis, the expectation is that this call works the same.

### Steps to Reproduce
Have the below code in your activity. Then,...
- turn location off, start the activity. The app will log _"The location settings are off, let's ask the user to turn them on!"_ and immediately show the location settings activity. However, the user is not led back to your activity after he turned it on. If he goes back manually (by pressing back), this is interpreted as `Activity.RESULT_CANCELED` and _"Oh no, user did not turn location on!"_ is logged. It seems to be unachievable to get a `Activity.RESULT_OK`
- turn location on, start the activity. The app will log _"Oh no, cannot change location settings!"_, even though they are already on. The status should have been `Status.SUCCESS` but is `Status.SETTINGS_CHANGE_UNAVAILABLE`
```
private static final int REQUEST_CHECK_SETTINGS = 1;

private LostApiClient lostApiClient;

@Override public void onStart()
{
	super.onStart();
	lostApiClient = new LostApiClient.Builder(this).build();
	lostApiClient.connect();
	requestLocationSettingsToBeOn();
}

@Override public void onStop()
{
	super.onStop();
	lostApiClient.disconnect();
}

private void requestLocationSettingsToBeOn()
{
	LocationRequest locationRequest = LocationRequest.create().setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY);
	PendingResult<LocationSettingsResult> result = LocationServices.SettingsApi.checkLocationSettings(
			lostApiClient, new LocationSettingsRequest.Builder().addLocationRequest(locationRequest).build());
	result.setResultCallback(new ResultCallback<LocationSettingsResult>()
	{
		@Override public void onResult(@NonNull LocationSettingsResult result)
		{
			final Status status = result.getStatus();
			switch (status.getStatusCode())
			{
				case Status.SUCCESS:
					Log.i("LOST","Location settings are on, continue!");
					break;
				case Status.RESOLUTION_REQUIRED:
					Log.i("LOST","The location settings are off, let's ask the user to turn them on!");
					status.startResolutionForResult(MainActivity.this, REQUEST_CHECK_SETTINGS);
					break;
				case Status.SETTINGS_CHANGE_UNAVAILABLE:
					Log.i("LOST","Oh no, cannot change location settings!");
					break;
			}
		}
	});
}

@Override protected void onActivityResult(int requestCode, int resultCode, Intent data)
{
	super.onActivityResult(requestCode, resultCode, data);
	switch (requestCode)
	{
		case REQUEST_CHECK_SETTINGS:
			switch (resultCode) {
				case Activity.RESULT_OK:
					Log.i("LOST","Hooray, user turned location on!");
					break;
				case Activity.RESULT_CANCELED:
					Log.i("LOST","Oh no, user did not turn location on!");
					break;
			}
			break;
	}
}
```
### Lost & Android Version
2.1.1 on Android 6

sarahlensing:
### Overview
This adds an intermediate step between initiation of a Status' resolution and firing of the device's location activity to give the user a chance to reject changing location settings.

### Proposed Changes
Creates a new class, `SettingsDialogDisplayer` to handle showing an alert to prompt the user to continue to the device's location settings screen.

screenshot_20161212-131802

Closes #140

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/api/Status.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/Status.java
@@ -1,5 +1,7 @@
 package com.mapzen.android.lost.api;
 
+import com.mapzen.android.lost.internal.DialogDisplayer;
+
 import android.app.Activity;
 import android.app.PendingIntent;
 import android.content.IntentSender;
@@ -19,13 +21,14 @@ public class Status implements Result {
 
   private final int statusCode;
   private final String statusMessage;
-  private final PendingIntent mPendingIntent;
+  private final PendingIntent pendingIntent;
+  private final DialogDisplayer dialogDisplayer;
 
-  public Status(int statusCode) {
-    this(statusCode, null);
+  public Status(int statusCode, DialogDisplayer dialogDisplayer) {
+    this(statusCode, dialogDisplayer, null);
   }
 
-  public Status(int statusCode, PendingIntent pendingIntent) {
+  public Status(int statusCode, DialogDisplayer dialogDisplayer, PendingIntent pendingIntent) {
     String statusMessage;
     switch (statusCode) {
       case SUCCESS:
@@ -55,7 +58,8 @@ public class Status implements Result {
     }
     this.statusCode = statusCode;
     this.statusMessage = statusMessage;
-    this.mPendingIntent = pendingIntent;
+    this.pendingIntent = pendingIntent;
+    this.dialogDisplayer = dialogDisplayer;
   }
 
   /**
@@ -67,11 +71,10 @@ public class Status implements Result {
    * @param requestCode associated with activity.
    * @throws IntentSender.SendIntentException
    */
-  public void startResolutionForResult(Activity activity, int requestCode)
+  public void startResolutionForResult(final Activity activity, final int requestCode)
       throws IntentSender.SendIntentException {
     if (this.hasResolution()) {
-      activity.startIntentSenderForResult(this.mPendingIntent.getIntentSender(), requestCode, null,
-          0, 0, 0);
+      dialogDisplayer.displayDialog(activity, requestCode, pendingIntent);
     }
   }
 
@@ -88,7 +91,7 @@ public class Status implements Result {
    * @return whether or not there is a resolution.
    */
   public boolean hasResolution() {
-    return this.mPendingIntent != null;
+    return this.pendingIntent != null;
   }
 
   /**
@@ -128,7 +131,7 @@ public class Status implements Result {
    * @return the {@link PendingIntent}.
    */
   public PendingIntent getResolution() {
-    return this.mPendingIntent;
+    return this.pendingIntent;
   }
 
   @Override public Status getStatus() {
--- a/lost/src/main/java/com/mapzen/android/lost/api/Status.java
+++ b/lost/src/main/java/com/mapzen/android/lost/api/Status.java
@@ -66,7 +66,11 @@ public class Status implements Result {
    * If the status code is {@link Status#RESOLUTION_REQUIRED}, then this method can be called to
    * start the resolution. For example, it will launch the Settings {@link Activity} so that the
    * user can update location settings when used with {@link SettingsApi#checkLocationSettings(
-   * LostApiClient, LocationSettingsRequest)}
+   * LostApiClient, LocationSettingsRequest)}. This activity will finish with
+   * {@link Activity#onActivityResult(int, int, android.content.Intent)} but the resultCode will
+   * never be {@code Activity#RESULT_OK}. You should instead rely on the requestCode to determine
+   * application flow and assume that the result is {@code Activity#RESULT_OK}.
+   *
    * @param activity to launch for resolution.
    * @param requestCode associated with activity.
    * @throws IntentSender.SendIntentException
--- a/lost/src/main/java/com/mapzen/android/lost/internal/DialogDisplayer.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/DialogDisplayer.java
@@ -1,0 +1,13 @@
+package com.mapzen.android.lost.internal;
+
+import android.app.Activity;
+import android.app.PendingIntent;
+
+/**
+ * Created by sarahlensing on 12/12/16.
+ */
+
+public interface DialogDisplayer {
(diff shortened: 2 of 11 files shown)


msmollin:
I go back and forth on having frameworks like this handle UI display. On the one hand, I can see the appeal as it centralizes management and keeps the external API cleaner. On the other hand, we then have to support UI components inside the SDK and it can get to be a mess when people want to customize all the things. Thoughts?

ecgreb:
@msmollin you make a good point. In this instance we decided to model the behavior after the [play services settings API](https://developers.google.com/android/reference/com/google/android/gms/location/SettingsApi) which also shows a dialog.

> If the status code is RESOLUTION_REQUIRED, the client can call startResolutionForResult(Activity, int) to bring up a dialog, asking for user's permission to modify the location settings to satisfy those requests.

What we could do is add a hook to set a custom `DialogDisplayer` which would allow us to provide the same level of convenience with a greater degree of control.

msmollin:
Agreed, and yes iOS is the same way - Core Location manages the display of that dialog. However it's also one of the things iOS developers complain about not being able to style / control the display, so I think having a custom hook is a good idea.

westnordost:
It is not so much about convenience but about the goal of LOST to be a drop-in replacement for the Google Play Api in that matter.
However, keep in mind that the workflow as it is in Google Play Api is broken for LOST anyway:
1. user must go into settings dialog and change it there
2. there will be no `Status.SUCCESS` return code when the user switched the setting
3. the user must go back to the app manually
This can't be fixed. So, the drop-in replacement does simply not work here.

Thus, if it can't work the same as Google Play Api anyway, the question is whether it makes sense to adhere to that Api for the resolution in function naming. A user that switches from Google Play Api to LOST will collide with it either way - whether the function is missing or not.

A convenience method to open up the location settings dialog and a thorough explanation of the alternative workflow for LOST in the javadoc could also be an option.

Calling startResolutionForResult, supplying your custom DialogDisplayer and correctly handling the "answer" would be I think about the same amount of code for the user than showing an alert, sending the user to the settings dialog and rechecking when the user comes back manually.

lost/src/main/java/com/mapzen/android/lost/api/Status.java
```diff
@@ -67,11 +71,10 @@ public Status(int statusCode, PendingIntent pendingIntent) {
    * @param requestCode associated with activity.
    * @throws IntentSender.SendIntentException
    */
-  public void startResolutionForResult(Activity activity, int requestCode)
+  public void startResolutionForResult(final Activity activity, final int requestCode)
```
ecgreb: Per comment by @westnordost we should update the Javadoc here to specifically call out the difference with play services in that currently `onActivityResult(...)` will not be invoked for the given `activity` parameter with `RESULT_OK`.

Another option would be to wrap the `SettingsDialogFragment` in its own `Activity` which could perform an additional location settings check when the user returns from the system settings menu. Then if the settings have been updated by the user it could finish itself with `RESULT_OK`.

Of course then we would also need a way to pass the original `LocationSettingsRequest` into the wrapper `Activity` so it could verify location settings requirements have not been met.

ecgreb: Perhaps the best course of action is to update the Javadoc for now and create an issue to implement the wrapper activity that can make an additional settings request and set `RESULT_OK` if conditions have been met at a later time.

lost/src/main/java/com/mapzen/android/lost/internal/DialogDisplayer.java
```diff
@@ -0,0 +1,13 @@
+package com.mapzen.android.lost.internal;
+
+import android.app.Activity;
+import android.app.PendingIntent;
+
+/**
+ * Created by sarahlensing on 12/12/16.
```
sarahlensing: Updated.

lost/src/main/java/com/mapzen/android/lost/internal/DialogDisplayer.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+
```
sarahlensing: Rm'd

lost/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="settings_alert_title">For best results let your device turn on location using
+  Mapzen\'s location service.\n\nClicking \"OK\" will bring you to the device\'s location settings screen.</string>
```
sarahlensing: Updated

lost/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,7 @@
...
+  <string name="cancel">Cancel</string>
+  <string name="ok">Ok</string>
+</resources>
```
sarahlensing: Added newline

lost/src/main/java/com/mapzen/android/lost/api/Status.java
sarahlensing: Updated docs and created issue: https://github.com/mapzen/lost/issues/151

PR #152, opened 2017-01-06
Extracts `LocationClientWrapper` class

ecgreb:
### Overview

Adds `LocationClientWrapper` class to encapsulate a single `LostApiClient` and all its associated listeners, pending intents, callbacks, and loopers.

### Proposed Changes

This change simplifies the logic in `LostClientManager` for adding and removing listeners and sending location updates.

As a second stage to this refactor we may want to consider migrating some of the logic from `LostClientManager.iterateAndNotify()` into the `ClientWrapper` class.


sarahlensing:
👍

PR #163, opened 2017-02-23
Fixes ConcurrentModificationException when client is disconnected on connect

### LINKED ISSUE #162: ConcurrentModificationException when disconnecting on onLocationChanged
I need a possibility to request **one** location update, so I wrote this class: [SingleLocationRequest.java.txt](https://github.com/mapzen/lost/files/786214/SingleLocationRequest.java.txt). Basically, when started, it requests location updates and on the first `onLocationChanged()` it receives, it disconnects again.

In the current LOST release 2.1.2, this can lead to the following exception
```
E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: de.westnordost.streetcomplete, PID: 32694
                  java.util.ConcurrentModificationException
                      at java.util.HashMap$HashIterator.nextEntry(HashMap.java:787)
                      at java.util.HashMap$KeyIterator.next(HashMap.java:814)
                      at com.mapzen.android.lost.internal.FusedLocationServiceConnectionManager.onServiceConnected(FusedLocationServiceConnectionManager.java:83)
                      at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl.onServiceConnected(FusedLocationProviderApiImpl.java:65)
                      at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1214)
                      at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1231)
                      at android.os.Handler.handleCallback(Handler.java:739)
                      at android.os.Looper.loop(Looper.java:211)
                      at android.app.ActivityThread.main(ActivityThread.java:5389)
                      at java.lang.reflect.Method.invoke(Native Method)
                      at java.lang.reflect.Method.invoke(Method.java:372)
                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1020)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:815)
```
I researched a bit as for the reason why and found the problem. See this figure of the stacktrace.
stracktrace
It is likely that it is only reproducible with at least 2 clients that connect at the same time.

Actually, a "working example" of this issue could be reduced: A client that disconnects again right on `onConnect()` will spark the same problem as the posted class of mine.

### Workaround
The workaround for users is to defer disconnecting from the lost api client, i.e.
```
Handler h = new Handler(Looper.getMainLooper());
h.post(new Runnable()
{
	@Override public void run()
	{
		LocationServices.FusedLocationApi.removeLocationUpdates(lostApiClient, this);
		lostApiClient.disconnect();
	}
});
```

### Solution
I think the most straightforward and easiest solution is to copy the `connectionCallbacks` in FusedLocationServiceConnectionManager and iterate on the copy in `onServiceConnected` and `onServiceDisconnected`.
Also, it might make sense to check this race condition for other callbacks that the lost api makes to its clients.

ecgreb:
### Overview

Fixes `ConcurrentModificationException` when multiple clients exist and one client is disconnected on connect.

### Proposed Changes

Iterates a over copy of the connection callbacks set stored in the `FusedLocationServiceConnectionManager` when the fused location service is either connected or suspended. This prevent a `ConcurrentModificationException` if callbacks are removed during iteration.

Fixes #162

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
@@ -6,6 +6,7 @@ import com.mapzen.android.lost.api.LostApiClient.ConnectionCallbacks;
 import android.content.Context;
 import android.os.IBinder;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -80,7 +81,7 @@ public class FusedLocationServiceConnectionManager {
       }
 
       if (!connectionCallbacks.isEmpty()) {
-        final Set<ConnectionCallbacks> copy = new HashSet<>(connectionCallbacks);
+        final ArrayList<ConnectionCallbacks> copy = new ArrayList<>(connectionCallbacks);
         for (LostApiClient.ConnectionCallbacks callbacks : copy) {
           callbacks.onConnected();
         }
@@ -92,7 +93,7 @@ public class FusedLocationServiceConnectionManager {
     if (connectState != ConnectState.IDLE) {
       connectState = ConnectState.IDLE;
       if (!connectionCallbacks.isEmpty()) {
-        final Set<ConnectionCallbacks> copy = new HashSet<>(connectionCallbacks);
+        final ArrayList<ConnectionCallbacks> copy = new ArrayList<>(connectionCallbacks);
         for (LostApiClient.ConnectionCallbacks callbacks : copy) {
           callbacks.onConnectionSuspended();
         }
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
@@ -80,7 +80,8 @@ public class FusedLocationServiceConnectionManager {
       }
 
       if (!connectionCallbacks.isEmpty()) {
-        for (LostApiClient.ConnectionCallbacks callbacks : connectionCallbacks) {
+        final Set<ConnectionCallbacks> copy = new HashSet<>(connectionCallbacks);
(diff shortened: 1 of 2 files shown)


westnordost:
nice, thank you :-)
Looking forward for the next LOST release, my fingers are itching to remove those workarounds

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
```diff
@@ -80,7 +80,8 @@ public void onServiceConnected(IBinder binder) {
       }
 
       if (!connectionCallbacks.isEmpty()) {
-        for (LostApiClient.ConnectionCallbacks callbacks : connectionCallbacks) {
+        final Set<ConnectionCallbacks> copy = new HashSet<>(connectionCallbacks);
```
westnordost: You could also copy it into an ArrayList to reserve less memory.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
```diff
@@ -91,7 +92,8 @@ public void onServiceDisconnected() {
     if (connectState != ConnectState.IDLE) {
       connectState = ConnectState.IDLE;
       if (!connectionCallbacks.isEmpty()) {
-        for (LostApiClient.ConnectionCallbacks callbacks : connectionCallbacks) {
+        final Set<ConnectionCallbacks> copy = new HashSet<>(connectionCallbacks);
```
westnordost: same

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceConnectionManager.java
ecgreb: Good catch - updated.

PR #174, opened 2017-03-24
GPS feature should not be required to use Lost

### LINKED ISSUE #169: gps required
Just built a map app using mapbox which uses this library. The app is built for android wear which often don't have gps sensors, this library uses:

<uses-feature android:name="android.hardware.location.gps"/>

Which adds required = true (as its not specified) when merging manifests. Searching the library it shows gpsPresent boolean so I'm guessing gps required could be set to false.

<uses-feature android:name="android.hardware.location.gps"  android:required="false"/>

Hopefully its an easy change

Craig

ecgreb:
### Overview

GPS sensor should not be required to use Lost.

### Proposed Changes

Updates `AndroidManifest.xml` to set `required=false` for `android.hardware.location.gps` feature.

Closes #169

### DIFF
--- a/lost/src/main/AndroidManifest.xml
+++ b/lost/src/main/AndroidManifest.xml
@@ -6,7 +6,9 @@
       android:name="android.hardware.sensor.accelerometer"
       android:required="false"/>
   <uses-feature android:name="android.hardware.location"/>
-  <uses-feature android:name="android.hardware.location.gps"/>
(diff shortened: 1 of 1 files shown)


sarahlensing:
Looks good. Maybe we want to do the same thing with the remaining, non-optional hardware requirements: https://github.com/mapzen/lost/issues/176

PR #180, opened 2017-04-05
Simplify service shutdown to prevent crash on screen rotation

### LINKED ISSUE #170: Crash on screen rotate with multiple lostApiClients
### Description
When I rotate the screen in my [app](https://github.com/westnordost/streetcomplete), LOST throws the following exception:
```
java.lang.IllegalArgumentException: Client must be added before requesting location updates
    at com.mapzen.android.lost.internal.LostClientManager.throwIfClientNotAdded(LostClientManager.java:98)
    at com.mapzen.android.lost.internal.LostClientManager.addListener(LostClientManager.java:76)
    at com.mapzen.android.lost.internal.FusedLocationProviderServiceImpl.requestLocationUpdates(FusedLocationProviderServiceImpl.java:59)
    at com.mapzen.android.lost.internal.FusedLocationProviderService.requestLocationUpdates(FusedLocationProviderService.java:69)
    at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl.requestLocationUpdates(FusedLocationProviderApiImpl.java:109)
    at de.westnordost.streetcomplete.tangram.MapFragment.onConnected(MapFragment.java:448)
    at com.mapzen.android.lost.internal.FusedLocationServiceConnectionManager.onServiceConnected(FusedLocationServiceConnectionManager.java:86)
    at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl.onServiceConnected(FusedLocationProviderApiImpl.java:65)
    at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1214)
    at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1231)
    at android.os.Handler.handleCallback(Handler.java:739)
    at android.os.Looper.loop(Looper.java:211)
    at android.app.ActivityThread.main(ActivityThread.java:5389)
    at java.lang.reflect.Method.invoke(Native Method)
    at java.lang.reflect.Method.invoke(Method.java:372)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1020)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:815)
```
### Steps to Reproduce
It is always reproducable within my app but unknown for a minimal sample project.
The usage of LOST in my app is a little more complex, since I use several LostApiClients from within different classes. Basically, each fragment that uses a LostApiClient makes the following calls in this order:

1. in `onAttach`, a LostApiClient is built using `new LostApiClient.Builder(context).addConnectionCallbacks(this).build()`
2. some time later (when another fragment ensured the user has all permissions and GPS is on), `startPositionTracking()` in the fragment is called, which executes `if(!lostApiClient.isConnected()) lostApiClient.connect();`
3. in `onConnected`, finally `LocationServices.FusedLocationApi.requestLocationUpdates(lostApiClient, request, this)` is called.
4. later in `onStop` or from outside, `stopPositionTracking()` is called in the fragment, which executes the following code: `LocationServices.FusedLocationApi.removeLocationUpdates(lostApiClient, this); lostApiClient.disconnect();`
5. when rotating the screen, 1. and 2. is called thereafter, LOST throws the exception in 3.

### Lost & Android Version
LOST 2.2.0, Android 5.1.1

ecgreb:
### Overview

Simplifies service shutdown process to prevent race condition that leads to a crash when the device is rotated.

### Proposed Changes

Removes `FusedLocationProviderServiceImpl#shutdown()`. This method was previously invoked when the system called `FusedLocationProviderService#onDestroy()` and did two things:

1. Clear all location requests by calling `LocationEngine#setRequest(null)`.
2. Clear all location clients from the client manager by calling `LostClientManager#shutdown()`.

The problem is that `FusedLocationProviderService#onDestroy()` would be invoked asynchronously by the system after rotating the device. By the time `onDestroy()` fires the client application may have already re-connected the location client and/or made a new request for location updates.

If `onDestroy()` fires sometime between connecting the client and requesting updates it would lead to a crash similar those reported in #170.

Fortunately it seems neither of the operations in `FusedLocationProviderServiceImpl#shutdown()` are needed since **both are already completed by the time `onDestroy()` is invoked by the system**:

1. Anytime a location request is removed `checkAllListenersPendingIntentsAndCallbacks()` is called which calls `LocationEngine#setRequest(null)` if all listeners have been removed (a prerequisite to stopping the service).
2. In `LostApiClientImpl#disconnect()` the fused location service is only shutdown if there are no remaining clients so clearing the clients again when the service is destroyed is unnecessary.

### Other Changes and Cleanup

* Removes public method `shutdown()` from `ClientManager` interface.
* Replaces with package private method `LostClientManager#clearClients()` visible for testing only.
* Adds `@Override` annotations to `ClientManager` interface methods implemented in `LostClientManager`.
* Shortens request intervals in `MultipleLocationListenerSingleClientActivity` to speed up feedback loop.

Fixes #170

### DIFF
--- a/lost-sample/src/main/java/com/example/lost/MultipleLocationListenerSingleClientActivity.java
+++ b/lost-sample/src/main/java/com/example/lost/MultipleLocationListenerSingleClientActivity.java
@@ -101,7 +101,7 @@ public class MultipleLocationListenerSingleClientActivity extends ListActivity {
       return;
     }
 
-    long interval = 3 * 60 * 1000; // 3 minutes
+    long interval = 30 * 1000; // 30 seconds
     LocationRequest request = LocationRequest.create()
         .setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)
         .setFastestInterval(interval)
@@ -109,7 +109,7 @@ public class MultipleLocationListenerSingleClientActivity extends ListActivity {
 
     LocationServices.FusedLocationApi.requestLocationUpdates(lostApiClient, request, listener);
 
-    interval = 30 * 1000; // 30 seconds
+    interval = 15 * 1000; // 15 seconds
     request = LocationRequest.create()
         .setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)
         .setFastestInterval(interval)
--- a/lost/src/main/java/com/mapzen/android/lost/internal/ClientManager.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/ClientManager.java
@@ -43,7 +43,6 @@ public interface ClientManager {
   void reportProviderDisabled(String provider);
   void notifyLocationAvailability(final LocationAvailability availability);
   boolean hasNoListeners();
-  void shutdown();
   Map<LostApiClient, Set<LocationListener>> getLocationListeners();
   Map<LostApiClient, Set<PendingIntent>> getPendingIntents();
   Map<LostApiClient, Set<LocationCallback>> getLocationCallbacks();
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderService.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderService.java
@@ -50,11 +50,6 @@ public class FusedLocationProviderService extends Service {
     serviceImpl = new FusedLocationProviderServiceImpl(this, LostClientManager.shared());
   }
 
-  @Override public void onDestroy() {
-    super.onDestroy();
-    serviceImpl.shutdown();
-  }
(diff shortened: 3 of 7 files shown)


ecgreb:
@westnordost can you see if this changes fixes the crash you reported in #170 (and doesn't introduce any side effects) since your stack trace was somewhat different from what we are able to reproduce with the Lost demo app?

ecgreb:
I built `StreetComplete` locally using a Lost 2.2.1-SNAPSHOT created from this branch and it does indeed seem to solve the crash introduced in version 2.2.0.

PR #208, opened 2017-05-23
Unregister callbacks

### LINKED ISSUE #206: Unregister ConnectionCallbacks
To prevent leaking resources we should provide a way to unregister callbacks added in `LostApiClient.Builder#addConnectionCallbacks`

sarahlensing:
### Overview
This PR adds `LostApiClient#unregisterConnectionCallbacks` so that developers can prevent leaking resources when adding callbacks in `LostApiClient.Builder.addConnectionCallbacks`

It also updates the sample app examples to use cleanup registered callbacks.

Future work will add `LostApiClient#registerConnectionCallbacks` to allow adding callbacks after the client has been created

Closes #206


lost/src/main/java/com/mapzen/android/lost/internal/GeofencingApiImpl.java
```diff
@@ -133,6 +133,7 @@ private void removeGeofences(LostApiClient client, String geofenceRequestId)
   @RequiresPermission(anyOf = {ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION})
   @Override public PendingResult<Status> removeGeofences(LostApiClient client,
       PendingIntent pendingIntent) throws SecurityException {
+
```
msmollin: Minor nit: remove if not needed.

PR #210, opened 2017-05-24
Throw IllegalStateException

### LINKED ISSUE #209: Throw detailed exception
We introduced a change with [this PR](https://github.com/mapzen/lost/pull/180) which simplifies what happens when a client disconnects. Before this PR, location updates would be removed on disconnect, but after, they no longer are.

To improve clarity for developers, we should provide specific information about why the exception is thrown and what they have done wrong. Previous to this change, developer misuse of the API would be transparent so long as they disconnected any `LostApiClient`s created because. Now, their apps will crash if they do not explicitly remove any requested location updates.

sarahlensing:
### Overview
This PR makes misuse of the APIs more clear and gives details about how to resolve the issue. It also refactors out the service callback code into a separate class and adds tests for this functionality (which was previously untested).

### Proposed Changes
We introduced a change with [this PR](https://github.com/mapzen/lost/pull/180) which simplifies what happens when a client disconnects. Before this PR, location updates would be removed on disconnect, but after, they no longer are. So, if a developer doesn't remove location updates, they will likely see a crash. Before this PR it would result in a NPE which was confusing because it implied there was an issue with LOST.

Closes #209

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -43,6 +43,12 @@ public class FusedLocationProviderApiImpl extends ApiImpl
     public void onLocationChanged(final Location location) throws RemoteException {
       new Handler(Looper.getMainLooper()).post(new Runnable() {
         @Override public void run() {
+          if (service == null) {
+            throw new IllegalStateException("Location update received after client was "
+                + "disconnected. Did you forget to unregister location updates before "
+                + "disconnecting?");
+          }
+
           final LostClientManager clientManager = LostClientManager.shared();
           ReportedChanges changes = clientManager.reportLocationChanged(location);
 
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -42,7 +42,7 @@ public class FusedLocationProviderApiImpl extends ApiImpl
     public void onLocationChanged(final Location location) throws RemoteException {
       new Handler(Looper.getMainLooper()).post(new Runnable() {
         @Override public void run() {
-          LostClientManager clientManager = LostClientManager.shared();
+          final LostClientManager clientManager = LostClientManager.shared();
           serviceCallbackManager.onLocationChanged(context, location, clientManager, service);
         }
       });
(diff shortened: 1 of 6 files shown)


tallytalwar:
On it.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -43,6 +43,12 @@
     public void onLocationChanged(final Location location) throws RemoteException {
       new Handler(Looper.getMainLooper()).post(new Runnable() {
         @Override public void run() {
+          if (service == null) {
+            throw new IllegalStateException("Location update received after client was "
```
tallytalwar: Can we also provide this information in the documentation for LOST?

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -43,34 +42,16 @@
...
-          final LocationResult result = LocationResult.create(locations);
-          ReportedChanges pendingIntentChanges = clientManager.sendPendingIntent(
-              context, location, availability, result);
-
-          ReportedChanges callbackChanges = clientManager.reportLocationResult(location, result);
-
-          changes.putAll(pendingIntentChanges);
-          changes.putAll(callbackChanges);
-          clientManager.updateReportedValues(changes);
+          LostClientManager clientManager = LostClientManager.shared();
```
tallytalwar: Why to remove `final` from here?

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceCallbackManager.java
```diff
@@ -0,0 +1,48 @@
...
+import android.os.RemoteException;
+
+import java.util.ArrayList;
+
+public class FusedLocationServiceCallbackManager {
+
+  void onLocationChanged(Context context, Location location, LostClientManager clientManager,
+      IFusedLocationProviderService service) {
+    if (service == null) {
+      throw new IllegalStateException("Location update received after client was "
```
tallytalwar: This should also be well documented in LOST documentation.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
sarahlensing: Whoops. Added it back.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationServiceCallbackManager.java
sarahlensing: Documented in `LostApiClient`

Issue #214, opened 2017-06-07
Simplify FusedLocationProvider Internal Check


sarahlensing:
Because `LocationEngine#removeRequest` disables and re-enables the engine (if there are still remaining requests), we no longer need to check to see if the engine should be disabled in `FusedLocationProviderApi#checkAllListenersPendingIntentsAndCallbacks` but as @ecgreb [pointed out](https://github.com/mapzen/lost/pull/215) we could as an enhancement call `LocationEngine#removeAllRequests` when the last client is disconnected "to avoid this issue of needing to explicitly remove all requests before shutting down a client"

msmollin:
Closing as this repo is no longer maintained.

PR #215, opened 2017-06-07
Update engine's location requests

### LINKED ISSUE #213: Location Provider keeps showing after HIGH_POWER updates are removed
### Description

Location provider icon in status bar keeps showing after removing HIGH_POWER location updates while using two lost api clients. 

### Steps to Reproduce

Have 2 components with each one lost api client. The first component is started when the application is created and requests NO_POWER updates as long as the application lives. The second component is started for a specific activity and requests HIGH_POWER updates. When that activity is destroyed we remove the location request but the location provider keeps showing. Only killing the application resolves this. 

LMK if you need more information.

### Lost & Android Version

LOST 3.0.0, Android API 25

sarahlensing:
### Overview
Before this PR, `LocationRequest`s were not removed from the underlying `LocationEngine`. This meant that the location requirements we requested from the system could sometimes be too aggressive. This PR adds behavior to remove all `LocationRequest`s for a given `LostApiClient` + `LocationListener`/`PendingIntent`/`LocationCallback` pair which are not used by any other pairs.

### Proposed Changes
- Creates new `LostRequestManager` class to manage which `LostApiClient` + `LocationListener`/`PendingIntent`/`LocationCallback` pairs have made which `LocationRequest`s
- Creates new class `ClientCallbackWrapper` to allow for a multi-key hash in `LostRequestManager`
- `FusedLocationProviderApi#requestLocationUpdates` has been updated to register `ClientCallbackWrapper`/`LocationRequest` pairs with the `LostRequestManager`
- `FusedLocationProviderApi#removeLocationUpdates` has been updated to unregister `ClientCallbackWrapper`/`LocationRequest` pairs with the `LostRequestManager`, retrieve all `LocationRequest`s used only by that `ClientCallbackWrapper`, and remove these `LocationRequest`s from the `LocationEngine`
- `LocationEngine#setRequest` has been removed and split into `LocationEngine#addRequest`, `LocationEngine#removeRequest`, and `LocationEngine#removeAllRequests`
- Because `LocationEngine#removeRequest` disables and re-enables the engine (if there are still remaining requests), we no longer need to check to see if the engine should be disabled in `FusedLocationProviderApi#checkAllListenersPendingIntentsAndCallbacks` ([ticket](https://github.com/mapzen/lost/issues/214) to remove this code in a follow up PR)

Followup work will [add more tests](https://github.com/mapzen/lost/issues/216)

Closes #213


tobrun:
Tested and resolves #213 :tada:

lost/src/main/java/com/mapzen/android/lost/internal/LocationEngine.java
```diff
@@ -35,21 +35,37 @@ public LocationEngine(Context context, Callback callback) {
...
+    if (request != null) {
+      this.request.removeRequest(request);
       disable();
+      if (!this.request.getRequests().isEmpty()) {
+        enable();
+      }
     }
   }
 
+  public void removeAllRequests() {
```
ecgreb: As a possible future enhancement, should this also be invoked when the last client is disconnected to avoid this issue of needing to explicitly remove all requests before shutting down a client?

lost/src/main/java/com/mapzen/android/lost/internal/LostRequestManager.java
```diff
@@ -0,0 +1,100 @@
...
+    Set<LocationRequest> requestsToRemove = clientCallbackToLocationRequests.get(wrapper);
+    if (requestsToRemove == null) {
+      return requestsToRemove;
+    }
+    clientCallbackToLocationRequests.remove(wrapper);
+    for (ClientCallbackWrapper w : clientCallbackToLocationRequests.keySet()) {
+      Set<LocationRequest> requests = clientCallbackToLocationRequests.get(w);
+      requestsToRemove.removeAll(requests);
+    }
+    return requestsToRemove;
```
ecgreb: I guess we can't rule out the possibility of more than one client registering the same request but that seems kind of gross.

What if a `LocationRequest` generated a unique internal id for itself based on the time it was registered? Would that simplify the logic here?

sarahlensing: I'm not sure this would simplify the logic because I'd still have to iterate over the request to unique id map to ensure the request id was the one corresponding to the client wrapper I wanted to remove requests for. I would also have to add a step to update request ids for remaining client wrappers after this step. Am I missing something though? Is this how you imagine it be used?

ecgreb: Mainly I was looking for a way to not have to search all other clients to see if another one is using any of the `requestsToRemove` before removing them but maybe a unique id is not the best way to achieve that.

What if we just made a copy of each incoming request so two clients could have equal requests without it being the same instance?

lost/src/main/java/com/mapzen/android/lost/internal/LocationEngine.java
sarahlensing: Good idea, we've run into some recent issues with this behavior. I updated this ticket to capture the work: https://github.com/mapzen/lost/issues/214

lost/src/main/java/com/mapzen/android/lost/internal/LostRequestManager.java
```diff
@@ -0,0 +1,100 @@
...
+      requests = new HashSet<>();
+      clientCallbackToLocationRequests.put(wrapper, requests);
+    }
+    requests.add(request);
+  }
+
+  private Set<LocationRequest> getRequestOnlyUsedBy(ClientCallbackWrapper wrapper) {
+    Set<LocationRequest> requestsToRemove = clientCallbackToLocationRequests.get(wrapper);
+    if (requestsToRemove == null) {
+      return requestsToRemove;
```
ecgreb: This could be simplified as `return null;`.

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -173,9 +178,24 @@ private void requestLocationUpdatesInternal(LocationRequest request) {
     }
   }
 
+  private void removeLocationUpdatesInternal(Set<LocationRequest> requests) {
+    if (requests == null) {
+      return;
+    }
+    for (LocationRequest request : requests) {
+      try {
+        service.removeLocationUpdates(request);
```
ecgreb: Sending a separate request to the service for each request to be removed will cause a disable/enable cycle for each one. This has the side effect of sending out multiple location updates in rapid succession via `FusionEngine#checkLastKnownAndNotify(provider)`.

I think active location listeners would be protected from this effect if they have a `fastestInterval` associate with their request but it might be worth considering a batch removal operation.

lost-sample/src/main/java/com/example/lost/MultiplePriorityMultipleClientsActivity.java
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+
+  private void requestNoPowerUpdates() {
+    LocationRequest request = LocationRequest.create()
+        .setPriority(LocationRequest.PRIORITY_NO_POWER)
+        .setFastestInterval(0)
+        .setSmallestDisplacement(0)
+        .setInterval(1000); // 1 sec
+    LocationServices.FusedLocationApi.requestLocationUpdates(client, request, noPowerListener);
```
sarahlensing: Nice catch, I'll create a new activity and migrate these examples in a future PR https://github.com/mapzen/lost/issues/217

zugaldia:
@sarahlensing anything that we could help with on our side to help getting this PR merged?

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -178,23 +179,21 @@ private void requestLocationUpdatesInternal(LocationRequest request) {
...
       return;
     }
-    for (LocationRequest request : requests) {
-      try {
-        service.removeLocationUpdates(request);
-      } catch (RemoteException e) {
-        throw new RuntimeException(e);
-      }
+    try {
+      service.removeLocationUpdates(requests);
```
ecgreb: This is great should really cut down on thrashing of the `LocationManager` listeners when multiple requests are removed!

PR #227, opened 2017-06-28
fix #224

### LINKED ISSUE #224: IllegalStateException on onLocationChanged
The following exception is reported by many of my users:
```
java.lang.IllegalStateException: 
  at com.mapzen.android.lost.internal.FusedLocationServiceCallbackManager.onLocationChanged(FusedLocationServiceCallbackManager.java:31)
  at com.mapzen.android.lost.internal.FusedLocationProviderApiImpl$1$1.run(FusedLocationProviderApiImpl.java:46)
  at android.os.Handler.handleCallback(Handler.java:751)
  at android.os.Looper.loop(Looper.java:154)
  at android.app.ActivityThread.main(ActivityThread.java:6077)
  at java.lang.reflect.Method.invoke(Native Method:0)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:865)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)

```

### Lost & Android Version
3.0.0
on Android 6.0, Android 7.0, Android 7.1
not on any version below Android 6.0 yet (but there are only 28 reports in total)

Unfortunately, I can not give any more details about it as I was not able to reproduce it myself yet. 

A helpful comment here: https://github.com/mapzen/lost/issues/218#issuecomment-310485055

westnordost:
The problem is at [FusedLocationProviderApiImpl line 48](https://github.com/mapzen/lost/blob/master/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java#L48). The null check must be done immediately on the calling thread, not deferred.

This fix was made blindly: I did not verify if the tests still run because the test setup does not work on my machine (I suppose because I am on Windows): `No such manifest file: build\intermediates\bundles\debug\--none`

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -39,13 +39,24 @@ public class FusedLocationProviderApiImpl extends ApiImpl
 
   IFusedLocationProviderService service;
 
-  private IFusedLocationProviderCallback.Stub remoteCallback
+  IFusedLocationProviderCallback.Stub remoteCallback
       = new IFusedLocationProviderCallback.Stub() {
     public void onLocationChanged(final Location location) throws RemoteException {
+
+      if (service == null) {
+        throw new IllegalStateException("Location update received after client was "
+                + "disconnected. Did you forget to unregister location updates before "
+                + "disconnecting?");
+      }
+
       new Handler(Looper.getMainLooper()).post(new Runnable() {
         @Override public void run() {
-          final LostClientManager clientManager = LostClientManager.shared();
-          serviceCallbackManager.onLocationChanged(context, location, clientManager, service);
+          // #224: this call is async, service may have been legally set to null in the meantime
+          if (service != null)
+          {
+            final LostClientManager clientManager = LostClientManager.shared();
+            serviceCallbackManager.onLocationChanged(context, location, clientManager, service);
+          }
         }
       });
     }
--- a/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
@@ -39,15 +39,9 @@ public class FusedLocationProviderApiImpl extends ApiImpl
 
   IFusedLocationProviderService service;
 
-  IFusedLocationProviderCallback.Stub remoteCallback
+  private IFusedLocationProviderCallback.Stub remoteCallback
       = new IFusedLocationProviderCallback.Stub() {
     public void onLocationChanged(final Location location) throws RemoteException {
-
-      if (service == null) {
-        throw new IllegalStateException("Location update received after client was "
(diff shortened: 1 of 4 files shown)


msmollin:
Hello and thanks for the PR! We'll definetly give this a spin and run the testing suite. 

We also ran into this issue with our own sample app, EraserMap over here: mapzen/eraser-map#824

We're in the process of testing out now whether or not Google Play Services behaves like LOST in this particular case, and if play services doesn't, we are going to probably wind up removing the the `IllegalStateException` all-together which will resolve this issue. Stay tuned!

westnordost:
Cool 👍

lost/src/main/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImpl.java
```diff
@@ -39,13 +39,24 @@
 
   IFusedLocationProviderService service;
 
-  private IFusedLocationProviderCallback.Stub remoteCallback
+  IFusedLocationProviderCallback.Stub remoteCallback
       = new IFusedLocationProviderCallback.Stub() {
     public void onLocationChanged(final Location location) throws RemoteException {
+
+      if (service == null) {
+        throw new IllegalStateException("Location update received after client was "
```
sarahlensing: I know its not your code but can you also remove this exception? Its too aggressive when compared to Play Services and silently ignoring the location update as you've done below is a fine solution.

lost/src/test/java/com/mapzen/android/lost/internal/FusedLocationProviderApiImplTest.java
```diff
@@ -154,6 +155,12 @@ public void requestLocationUpdates_callback_shouldThrowIfNotConnected() throws E
         new TestLocationCallback(), Looper.myLooper());
   }
 
+  @Test(expected = IllegalStateException.class)
```
sarahlensing: This can be removed too.

westnordost:
Sure. I have the same opinion on that.

On 28 June 2017 17:18:47 CEST, Sarah Lensing <notifications@github.com> wrote:
>sarahlensing commented on this pull request.
>
>Thanks for the PR! I just have one suggestion and then we can get this
>merged.
>
>>        = new IFusedLocationProviderCallback.Stub() {
>public void onLocationChanged(final Location location) throws
>RemoteException {
>+
>+      if (service == null) {
>+        throw new IllegalStateException("Location update received
>after client was "
>
>I know its not your code but can you also remove this exception? Its
>too aggressive when compared to Play Services and silently ignoring the
>location update as you've done below is a fine solution.
>
>> @@ -154,6 +155,12 @@ public void
>requestLocationUpdates_callback_shouldThrowIfNotConnected() throws E
>         new TestLocationCallback(), Looper.myLooper());
>   }
> 
>+  @Test(expected = IllegalStateException.class)
>
>This can be removed too.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/mapzen/lost/pull/227#pullrequestreview-46872218

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

PR #234, opened 2017-07-07
Shutdown Client Manager

### LINKED ISSUE #233: No location update after reconnect (LOST 3.0.2)
### Description
After stopping location updates (through tabbing out / changing Activity) and disconnecting from LOST api client, when it is attempted to reconnect later and request location updates anew, onLocationChanged is never called.

### Steps to Reproduce
For StreetComplete: Turn on location. Tab out or go to settings activity, then go back. The GPS button does not stop blinking (=app is waiting for a location update but does not get one).

**Program flow**:
1. `MainActivity.onStart()`: `locationRequestFragment.startRequest();` is called which ensures that the app has all the permissions and location is enabled. It then calls
2. `MainActivity.onLocationIsEnabled()`: The GPS button is made to blink. `singleLocationRequest.startRequest` ([source](https://github.com/westnordost/StreetComplete/blob/master/app/src/main/java/de/westnordost/streetcomplete/location/SingleLocationRequest.java)) connects to LOST, on connection requests location updates and then would, on receiving the first location update, make the GPS button stop blinking. The location update never comes.

**Edit**: Now after 5-10 minutes, the location update was received.

### Lost & Android Version
LOST 3.0.2

sarahlensing:
### Overview
This fixes a bug which prevented initial location updates from consistently being sent when clients re-connect to Lost.

### Proposed Changes
The `ClientManager` is a shared singleton used by `LostApiClient` and `FusedLocationProviderApi` to report location updates. When the service is started for the first  time, the `ClientManager` doesn't have any `ReportedChanges` and always sends the first location received to registered listeners. When the service is stopped and restarted quickly, the last known location reported by the service may not have changed and because the `ClientManager`'s `ReportedChanges` were not cleared, the location reported might not be sent to registered listeners (because it was compared to "stale" `ReportedChanges`). This PR adds code to "shutdown" the `ClientManager` when the service disconnects so that expected initial location updates are always sent when clients re-connect.

Small improvements (constructor injection) were made to the `FusedLocationProviderApiImpl` to make testing easier.

Closes #233

### DIFF
--- a/lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
@@ -24,7 +24,7 @@ import static com.mapzen.android.lost.api.FusedLocationProviderApi.KEY_LOCATION_
 
 /**
  * Used by {@link LostApiClientImpl} to manage connected clients and by
- * {@link FusedLocationProviderServiceDelegate} to manage client's {@link LocationListener}s,
+ * {@link FusedLocationProviderApiImpl} to manage client's {@link LocationListener}s,
  * {@link PendingIntent}s, and {@link LocationCallback}s.
  */
 public class LostClientManager implements ClientManager {
--- a/lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
+++ b/lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
@@ -212,6 +212,10 @@ public class LostClientManager implements ClientManager {
     return true;
   }
 
+  @Override public void shutdown() {
+    reportedChanges.clearAll();
+  }
+
   @VisibleForTesting void clearClients() {
     clients.clear();
   }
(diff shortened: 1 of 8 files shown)


lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
```diff
@@ -212,6 +212,10 @@ private void throwIfClientNotAdded(LostApiClient client) {
     return true;
   }
 
+  @Override public void shutdown() {
```
tallytalwar: Is `shutdown` an apt naming for this function? How about `clearReportedLocations` or something on those lines?

msmollin: If I'm recalling this change correctly, we landed on `shutdown` to make it a little more clear at the call site what this function's intended job is, with the intent to add future functionality. @sarahlensing might remember more.

tallytalwar: K, maybe good to merge to get the fix in and name resolution can be done later.

westnordost:
Do you have an ETA when you will release a patch with this fix?

sarahlensing:
We are going to cut a patch release by EOD today

lost/src/main/java/com/mapzen/android/lost/internal/LostClientManager.java
sarahlensing: Missed this comment earlier...what @msmollin mentions is correct though!

westnordost:
Nice :-)

msmollin:
Release has been cut: https://github.com/mapzen/lost/releases/tag/lost-3.0.3. I think we're mid deploy to MavenCentral but it all should get up there in an hour or so.

PR #246, opened 2017-10-16
Improve Circle builds

sarahlensing:
### Overview
Limits jvm heap to prevent build failures on circle

### Proposed Changes
https://circleci.com/docs/1.0/oom/#out-of-memory-errors-in-android-builds

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   java:
     version: oraclejdk8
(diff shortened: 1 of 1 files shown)


tallytalwar:
Interesting, I thought we had this, a while back I used to run into this issue when building eraser-map locally. But make sense merging.

PR #263, opened 2018-07-12
fix typo (minor)

serv-inc:
"to for" use

### Overview
just fix the thing

### Proposed Changes
fix minor typo


serv-inc:
Thanks for the API, that is just awesome!

PR #270, opened 2021-07-22
Update android 12

Qubiz:
### Overview

Android 12 (SDK 31) requires that when you are using intent-filters the `android:exported` attribute should be provided explicitly. Reference: https://developer.android.com/guide/topics/manifest/activity-element#exported

### Proposed Changes

Add the `android:exported="false"` attribute to the service(s) that use intent-filters.

### DIFF
--- a/lost/src/main/AndroidManifest.xml
+++ b/lost/src/main/AndroidManifest.xml
@@ -19,7 +19,9 @@
     <service
         android:name="com.mapzen.android.lost.internal.FusedLocationProviderService"
         android:process=":lost"/>
-    <service android:name="com.mapzen.android.lost.internal.GeofencingIntentService">
+    <service 
+        android:name="com.mapzen.android.lost.internal.GeofencingIntentService"
+        android:exported="true">
       <intent-filter>
         <action android:name="com.mapzen.lost.action.ACTION_GEOFENCING_SERVICE"/>
       </intent-filter>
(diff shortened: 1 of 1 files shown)


Eyadoos:
What's the status of this PR? I can't update my project to Android SDK 33 because of this error and it seems the library is completely abandoned since 2017, any idea how to solve this? I don't use the mapzen library directly, but it seems one of the libraries is using it internally.

Qubiz:
> What's the status of this PR? I can't update my project to Android SDK 33 because of this error and it seems the library is completely abandoned since 2017, any idea how to solve this? I don't use the mapzen library directly, but it seems one of the libraries is using it internally.

We had a similar issue. The workaround we have is to add the following service to our application's manifest:

```xml  
<!--
    This service should normally not be declared here. This is added here because
    Android 12 (SDK 31) requires explicit 'android:exported' attribute for services and
    activities that provide an intent-filter. This library (https://github.com/lostzen/lost)
    has not yet been updated to support this so we override it here.
-->
<service
    android:name="com.mapzen.android.lost.internal.GeofencingIntentService"
    android:exported="false">
    <intent-filter>
        <action android:name="com.mapzen.lost.action.ACTION_GEOFENCING_SERVICE" />
    </intent-filter>
</service>
```

msmollin:
Closing as this repo is no longer maintained.

